PR #8, opened 2020-08-03
ci: reduce complexity of build pipeline

blacha:
Everything currently supports python 3.8
Join test/lint job together

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,18 +1,15 @@
 name: Build
-on: [push, pull_request]
+on: [push]
 
 jobs:
-  linter:
+  build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python: ["3.6", "3.7", "3.8"]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,10 +18,6 @@ jobs:
         if: github.event_name == 'push'
         run: |
           gitlint
-      - name: Lint All Commits on Pull Request
```
dwsilk, line 24: Removing this will mean `gitlint` only checks the last commit, not all commits on a PR. I guess if kodiak is always used / configured to squash and merge, this is okay?

dwsilk: Relevant: https://github.com/linz/template-python-hello-world/issues/36

blacha: this takes out the `pull_request` event, I'm not sure we want to have both build and `pull_request` as everything will double up if you push to this repo and make a PR.

dwsilk (approved):

Issue #81, opened 2020-11-18
CSV Extractor is too slow


MDavidson17:
### Bug Description
Currently, when running the code for each file the CSV is opened and read. This is time-consuming.  
Instead, I want to preprocess the CSV (turn it into a dictionary and store in memory)

#### Steps to Reproduce
1. run code

#### Desktop
 - Environment: linux

### Note
- Also want to have a session to explain algorthimic complexity

PR #86, opened 2020-11-22
fix: speed up csv extractor

Ben-Cheng:
Fixes: #81

### DIFF
--- a/topo_processor/metadata/loaders/metadata_loader_imagery_historic.py
+++ b/topo_processor/metadata/loaders/metadata_loader_imagery_historic.py
@@ -1,5 +1,6 @@
 import csv
 import os
+from typing import Dict
 
 from topo_processor.metadata.data_type import DataType
 from topo_processor.metadata.item import Item
@@ -10,6 +11,8 @@ from .metadata_loader import MetadataLoader
 
 class MetadataLoaderImageryHistoric(MetadataLoader):
     name = "loader.imagery.historic"
+    is_init = False
+    csv_dict: Dict[str, Dict[str, str]] = {}
 
     def is_applicable(self, item: Item) -> bool:
         if item.collection.data_type != DataType.ImageryHistoric:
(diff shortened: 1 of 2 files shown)


topo_processor/metadata/loaders/metadata_loader_imagery_historic.py
```diff
@@ -10,6 +11,9 @@
 
 class MetadataLoaderImageryHistoric(MetadataLoader):
     name = "loader.imagery.historic"
+    csv_dict = {}
```
blacha, line 14: should type the dictionary

topo_processor/metadata/loaders/metadata_loader_imagery_historic.py
```diff
@@ -19,37 +23,39 @@ def is_applicable(self, item: Item) -> bool:
         return True
 
     def add_metadata(self, item: Item) -> None:
-        file_name = os.path.splitext(os.path.basename(item.path))[0]
+        if not self.is_init:
+            self.read_csv()
+        item_basename = os.path.splitext(os.path.basename(item.path))[0]
+        item_metadata = self.csv_dict[item_basename]
+        properties = {
+            "linz:sufi": item_metadata[self.csv_header.index("sufi")],
```
blacha, line 31: is there a way to store these indexes and not re look them up for every line

Ben-Cheng: good idea, we will look into that

Ben-Cheng: investigate parse as dictionary / array in the csv lib

topo_processor/metadata/loaders/metadata_loader_imagery_historic.py
```diff
@@ -19,37 +22,39 @@ def is_applicable(self, item: Item) -> bool:
         return True
 
     def add_metadata(self, item: Item) -> None:
-        file_name = os.path.splitext(os.path.basename(item.path))[0]
+        if not self.is_init:
+            self.read_csv()
+
+        item_path_basename = os.path.splitext(os.path.basename(item.path))[0]
+        item_dict = self.csv_dict[item_path_basename]
```
blacha, line 29: `item_dict` may be null if the file does not exist.

Ben-Cheng: fixed
```
        if item_path_basename not in self.csv_dict:
            raise Exception(f"{item_path_basename} cannot be found in the csv.")
```

blacha (approved):

PR #87, opened 2020-11-24
refactor: made logger clearer

Ben-Cheng:
### Change Description:
Before:
image

After:
image

### DIFF
--- a/topo_processor/metadata/collection_factory.py
+++ b/topo_processor/metadata/collection_factory.py
@@ -13,16 +13,16 @@ from .item import Item
 def create_collection(path: str, data_type: DataType) -> Collection:
     start_time = time_in_ms()
     collection = Collection("title", "description", "license", data_type)
-    create_item(collection, path)
-    get_log().debug("CreateCollection", title=collection.title, data_type=data_type, duration=time_in_ms() - start_time)
+    create_items(collection, path)
+    get_log().debug("Collection loaded", title=collection.title, data_type=data_type, duration=time_in_ms() - start_time)
     return collection
 
 
-def create_item(collection: Collection, path: str) -> None:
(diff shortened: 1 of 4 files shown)


topo_processor/metadata/collection_factory.py
```diff
@@ -13,16 +13,14 @@
...
 
-def create_item(collection: Collection, path: str) -> None:
-    start_time = time_in_ms()
+def create_items(collection: Collection, path: str) -> None:
     for f in os.listdir(path):
         file_path = os.path.join(path, f)
         item = Item(file_path, collection)
         repo.add_metadata(item)
         collection.items.append(item)
-    get_log().debug("CreateItem", path=path, duration=time_in_ms() - start_time)
```
blacha, line 28: the total time to add metadata is quite a useful log to have, why was this taken out?

topo_processor/metadata/item.py
```diff
@@ -26,3 +27,4 @@ def __init__(self, path: str, collection: Collection):
             properties={},
             stac_extensions=[],
         )
+        get_log().info("Item initialised", path=self.path)
```
blacha, line 30: Once its been created its not actually initalised so this log is a little  missleading, I generally would avoid having logs in constructors too.

Ben-Cheng: why would you avoid having logs in constructors?

topo_processor/metadata/collection_factory.py
Ben-Cheng: I have added it back. I think I removed it by mistake.

Ben-Cheng:
Manually merged into the current feat/extract-info-from-tiff branch to avoid conflict

PR #119, opened 2021-02-23
feat: only cog historical imagery

Ben-Cheng:


topo_processor/cli/main.py
```diff
@@ -6,7 +6,7 @@
 import click
 from linz_logger import get_log
 
-from topo_processor.metadata import DataType, create_collection
+from topo_processor.factory import DataType, create_collection
```
blacha, line 9: factory is super generic name when its creating metadata why the change?

topo_processor/imagery/compressors/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .data_compressor_imagery_historic import DataCompressorImageryHistoric
```
blacha, line 1: might be better to call this a `transformer` and the have a compressor as type of a transformer

topo_processor/cli/main.py
Ben-Cheng: because it also calls compress data / tiffs in create_collection

topo_processor/imagery/compressors/__init__.py
Ben-Cheng: why is that? are we going to have other types of transformer in the future for lidar and aerial imagery?

blacha: Yes there will be more transformations later, cropping/filtering also a gdal_translate is a transformation, you don't actually have to compress a file to make a COG.

Ben-Cheng: yup will change that

topo_processor/cli/main.py
Ben-Cheng: I have renamed it to stac_object because they are mainly for creating collections and items. Thoughts? @blacha

Ben-Cheng: have renamed it to stac

blacha (approved):

PR #126, opened 2021-02-26
transformed tiffs and json files are stored in the same tmp folder

MDavidson17:
### Change Description:
transformed tiffs and json files are stored in the same tmp folder.
note: json files are only stored here for the s3 upload, they are stored in the target folder for the local upload

### DIFF
--- a/topo_processor/cog/command.py
+++ b/topo_processor/cog/command.py
@@ -23,11 +23,11 @@ class Command:
             self.container = docker_ref.get("container", None)
             self.container_tag = docker_ref.get("tag", None)
 
-    def append_arg(self, *args: str) -> "Command":
+    def append_arg(self, *args: str):
         for argument in args:
             self.arguments.append(argument)
 
-    def mount(self, *args: str) -> "Command":
+    def mount(self, *args: str):
         """Mount a folder, useful only if the command is run inside of docker"""
         for volume in args:
             self.volumes.append(volume)
--- a/topo_processor/cog/command.py
+++ b/topo_processor/cog/command.py
@@ -23,14 +23,16 @@ class Command:
             self.container = docker_ref.get("container", None)
             self.container_tag = docker_ref.get("tag", None)
 
-    def append_arg(self, *args: str):
+    def arg(self, *args: str) -> "Command":
         for argument in args:
             self.arguments.append(argument)
+        return self
 
-    def mount(self, *args: str):
+    def mount(self, *args: str) -> "Command":
         """Mount a folder, useful only if the command is run inside of docker"""
         for volume in args:
             self.volumes.append(volume)
+        return self
 
     def to_full_command(self) -> List[str]:
         return [self.command] + self.arguments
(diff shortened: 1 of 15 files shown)


MDavidson17:
also note: we have not changed the item asset reference, we will do this in a separate PR

topo_processor/stac/item.py
```diff
@@ -11,12 +11,14 @@ class Item:
...
     asset_basename: str
     asset_extension: str
+    transformed_asset_extension: str
     content_type: str
     collection: Collection
     stac_item: stac.Item
 
     def __init__(self, path: str, collection: Collection):
         self.path = path
+        self.transformed_data_path = None
```
blacha, line 21: what happens when there is more than one transformation?

topo_processor/cog/command.py
```diff
@@ -23,11 +23,11 @@ def __init__(self, command: str, docker_ref: CommandDocker = None):
             self.container = docker_ref.get("container", None)
             self.container_tag = docker_ref.get("tag", None)
 
-    def append_arg(self, *args: str) -> "Command":
+    def append_arg(self, *args: str):
```
blacha, line 26: these should `return self` and have the types left on.

topo_processor/cog/create_cog.py
```diff
@@ -3,10 +3,11 @@
 from topo_processor.cog.command import Command
 
 
-async def create_cog(input_file: str, output_dir: str, compression_method: str):
+async def create_cog(input_file: str, output_dir: str, compression_method: str) -> str:
 
-    gdal_translate = to_gdal_command(input_file, output_dir, compression_method)
+    gdal_translate, output_file = to_gdal_command(input_file, output_dir, compression_method)
```
blacha, line 8: should specify the output file and not let this command make one.

topo_processor/stac/collection.py
```diff
@@ -15,12 +16,14 @@ class Collection:
...
     license: str
     data_type: DataType
+    temp_dir: str
     items: List["Item"]
     providers: List[stac.Provider]
     collection_output_path: str
 
     def __init__(self, data_type: DataType):
         self.data_type = data_type
+        self.temp_dir = mkdtemp()
```
blacha, line 26: I don't think this logic should live here, this logic is around the collection not around temporary processing of it.

topo_processor/cog/command.py
Ben-Cheng: why does it need to return self?

topo_processor/stac/item.py
Ben-Cheng: it depends on where the other transformation happens? Why do we need more than one transformation in one item?

topo_processor/cog/command.py
blacha: It's so you can do something like


```
Command("foo").arg(bar).arg(baz).run()
```

topo_processor/stac/item.py
blacha: There will be lots of transformers one example is cog.

One example could be a tiff metadata adder, that adds or updates tiff metadata.

Could also apply gzip compression to a tiff after it's made

topo_processor/cog/command.py
Ben-Cheng: i see

topo_processor/stac/item.py
Ben-Cheng: i see. then this will overwrites the `transformed_data_path`. should we change it into a list of transformed_data_path?

Ben-Cheng (approved):

topo_processor/stac/item.py
Ben-Cheng: as discussed we will leave it for later because it requires some restructurings

PR #133, opened 2021-03-04
feat: hash data after data transformation

Ben-Cheng:


topo_processor/stac/item.py
```diff
@@ -7,18 +7,18 @@
...
-    item_output_path: str
-    asset_basename: str
-    asset_extension: str
-    transformed_asset_extension: str
+    path: str  # path to data file (input or transformed)
+    item_output_path: str  # destination path for stac item json file
+    asset_basename: str  # base name of asset (survey/sufi)
+    asset_extension: str  # extension of asset (lzw.cog.tiff)
     content_type: str
+    image_width: float
```
blacha, line 15: A stac item is not always a image, it may not have a width and height

topo_processor/stac/asset.py
```diff
@@ -0,0 +1,22 @@
...
+        item.stac_item.stac_extensions.append("file")
+
+    checksum_ = await multihash_as_hex(item.path)
+    item.stac_item.add_asset(
+        key="image",
+        asset=stac.Asset(
+            href=f"{item.asset_basename}.{item.asset_extension}",
+            properties={
+                "file:checksum": checksum_,
+                "linz:image_width": item.image_width,
```
blacha, line 17: maybe just drop the width/height for now

topo_processor/stac/asset.py
```diff
@@ -0,0 +1,22 @@
+import pystac as stac
+
+from topo_processor.util import multihash_as_hex
+
+
+async def add_asset(item):
```
blacha, line 6: might be better called add_asset_image, since your adding a image asset

MDavidson17 (approved):

PR #135, opened 2021-03-07
feat: implement stac collection

Ben-Cheng:


topo_processor/metadata/metadata_validators/metadata_validator_imagery_historic.py
```diff
@@ -0,0 +1,21 @@
...
+    def is_applicable(self, item: Item) -> bool:
+        if item.data_type != DataType.ImageryHistoric:
+            return False
+        if not is_tiff(item.source_file):
+            return False
+        return True
+
+    async def validate_metadata(self, item: Item) -> None:
+        item.is_valid = True
+        item.output_dir = item.stac_item.properties["linz:survey"]
```
blacha, line 21: this would be the prime spot to add the collection.json reference

Ben-Cheng: that will mean we will have to pass collection to this function and change the whole api

PR #139, opened 2021-03-10
refactor: renamed variables to make them more consistent across the repo

Ben-Cheng:


topo_processor/cog/create_cog.py
```diff
@@ -3,23 +3,21 @@
 from topo_processor.cog.command import Command
 
 
-async def create_cog(input_file: str, output_dir: str, compression_method: str) -> str:
-    output_file = os.path.join(output_dir, f"{os.path.basename(input_file)}.{compression_method}.cog.tiff")
-
-    gdal_translate = to_gdal_command(input_file, output_file, compression_method)
+async def create_cog(input_path: str, output_path: str, compression_method: str) -> str:
```
blacha, line 6: these two functions can be merged into one.

`to_gdal_command` is not a `to_gdal_command` as it has logic very specific to cog creation..

Ben-Cheng: These functions were separated for testing purpose. In create_cog_test.py, we have a function that tests the command has the right parameters.

blacha (approved):

PR #145, opened 2021-03-15
feat: handle multiple assets with transformations

Ben-Cheng:


topo_processor/data/data_transformers/data_transformer_imagery_historic.py
```diff
@@ -28,13 +29,15 @@ async def transform_data(self, item: Item) -> None:
...
-            "stac_extensions": ["file"],
-            "content_type": stac.MediaType.COG,
-        }
-        item.add_asset(asset)
+        item.add_asset(
+            "cog",
+            Asset(
+                key="image",
+                path=output_path,
+                properties={"file:checksum": None},
```
blacha, line 37: do you need to add the properties if there are none?

topo_processor/stac/item.py
```diff
@@ -1,63 +1,66 @@
...
+        self.content_type = pystac.MediaType.JSON
+        self.file_ext = MimeTypes().guess_extension(self.content_type)
         self.is_valid = True
+        self.assets = {
+            "source": Asset(
+                key="source",
+                path=source_path,
+                properties={"file:checksum": None},
+                content_type=MimeTypes().guess_type(source_path)[0],
+                file_ext=os.path.splitext(source_path)[1],
```
blacha, line 41: will this error without extensions?

topo_processor/stac/item.py
```diff
@@ -1,63 +1,66 @@
...
-        for asset in self.assets:
-            stac_item.add_asset(
-                key=asset["key"],
-                asset=stac.Asset(href=asset["href"], properties=asset["properties"], media_type=asset["content_type"]),
-            )
-        return stac_item
+        for asset_descriptor in self.assets:
+            asset = self.assets[asset_descriptor]
+            if asset.needs_upload:
+                asset.href = f"./{self.collection.title}/{self.id}{asset.file_ext}"
```
blacha, line 61: might want to validate that the href doesn't conflict with another asset.

topo_processor/stac/item.py
Ben-Cheng: yes, I will change this

topo_processor/data/data_transformers/data_transformer_imagery_historic.py
Ben-Cheng: no, properties is optional in stac asset. But I assumed we eventually will hash every asset so it will never be None?

Ben-Cheng: no, I have removed it now

topo_processor/stac/item.py
Ben-Cheng: no it will just return an empty string

topo_processor/stac/item.py
Ben-Cheng: added code to do this

Ben-Cheng:
@blacha ready for you to review again

topo_processor/stac/item.py
```diff
@@ -1,63 +1,70 @@
...
             properties=self.properties,
             stac_extensions=self.stac_extensions,
         )
-        for asset in self.assets:
-            stac_item.add_asset(
-                key=asset["key"],
-                asset=stac.Asset(href=asset["href"], properties=asset["properties"], media_type=asset["content_type"]),
-            )
-        return stac_item
+        existing_asset_hrefs = []
```
blacha, line 57: Don't use arrays for this, they are o(n) for lookups use a dict

Ben-Cheng: done

MDavidson17 (approved):

PR #147, opened 2021-03-16
feat: create collection store 

MDavidson17:


topo_processor/stac/collection_store.py
```diff
@@ -0,0 +1,14 @@
+from .collection import Collection
+
+
+class CollectionStore:
+    collections = {}
```
blacha, line 5: You either need to make these static, or make a single instance of the collection store.

Otherwise everytime you do `CollectionStore()` you get a whole new collection store

topo_processor/stac/tests/item_test.py
```diff
@@ -32,5 +30,6 @@ def test_duplicate_asset():
             needs_upload=True,
         ),
     )
-    with pytest.raises(Exception, match=r"./collection_title/item_id.tiff already exists."):
+    item.assets["source"].needs_upload = False
+    with pytest.raises(Exception, match=r"."):
```
blacha, line 34: matching any exception is not a good idea for tests

topo_processor/stac/collection_store.py
Ben-Cheng: done making it static

topo_processor/stac/tests/item_test.py
Ben-Cheng: changed back to with pytest.raises(Exception, match=r"./collection_title/item_id.tiff already exists."):

blacha (approved):

PR #153, opened 2021-03-19
feat: link collection and item stac objects

MDavidson17:


topo_processor/cog/execution.py
```diff
@@ -14,16 +10,12 @@ class ExecutionLocal:
 
     @staticmethod
     async def run(cmd: "Command"):
-        start_time = time_in_ms()
```
blacha, line 17: It is a very good idea to have this for when things go wrong, if it was too verbose, maybe switch it down to `trace` level

topo_processor/data/data_transformers/data_transformer_repo.py
```diff
@@ -21,8 +21,15 @@ async def transform_data(self, item: Item) -> None:
             for transformers in self.transformers:
                 if transformers.is_applicable(item):
                     start_time = time_in_ms()
-                    await transformers.transform_data(item)
+                    try:
+                        await transformers.transform_data(item)
+                    except Exception as error_msg:
+                        item.is_valid = False
```
blacha, line 27: you should return here, no point continuing.

topo_processor/uploader/local.py
```diff
@@ -11,13 +11,16 @@ async def upload_to_local_disk(collection: Collection, target: str):
...
                     asset.properties["file:checksum"] = await multihash_as_hex(asset.path)
                     copyfile(asset.path, os.path.join(target, item.parent, f"{item.id}{asset.file_ext}"))
-            await write_stac_metadata(item, os.path.join(target, item.parent, f"{item.id}{item.file_ext}"))
+                    stac_item = item.create_stac()
+                    stac_collection.add_item(stac_item)
+            await write_stac_metadata(stac_item, os.path.join(target, item.parent, f"{item.id}{item.file_ext}"))
         else:
-            get_log().info("Invalid item was not uploaded", source_path=item.source_path, error="TODO")  # TODO
-    await write_stac_metadata(collection, os.path.join(target, collection.title, f"collection{collection.file_ext}"))
+            get_log().warning("Invalid item was not uploaded:", error="; ".join(item.error_msgs), source_path=item.source_path)
```
blacha, line 25: you should be able to log a array, do you need to join the messages with `;`?

topo_processor/stac/item_factory.py
```diff
@@ -23,12 +23,14 @@ async def create_items(source_dir: str, data_type: DataType, target: str, temp_d
...
     if item.is_valid:
         await metadata_validator_repo.validate_metadata(item)
     if item.is_valid:
         await data_transformer_repo.transform_data(item)
     collection = get_collection(item.parent)
-    item.collection = collection
-    item.collection.metadata_path = os.path.join(item.parent, f"collection{collection.file_ext}")
-    collection.items.append(item)
+    if item.id in collection.items:
+        get_log().info("Duplicate Items", source_path=item.source_path)
```
blacha, line 33: Is duplicate really only a info log? should this be a exception?

topo_processor/data/data_transformers/data_transformer_repo.py
Ben-Cheng: done

topo_processor/uploader/local.py
Ben-Cheng: This is how it looks like if we log the array:
[2021-03-21T08:31:52.412Z] WARN Invalid item was not uploaded: error={ 0=this is an error 1=this is another error } source_path=/home/bcheng/dev/topo-processor/test_data/tiffs/399/CROWN_399_E_51.tiff

This is if we join them:
[2021-03-21T08:32:56.619Z] WARN Invalid item was not uploaded: error=this is an error; this is another error source_path=/home/bcheng/dev/topo-processor/test_data/tiffs/399/CROWN_399_E_51.tiff

Ben-Cheng: we will just log the array without joining them

topo_processor/stac/item_factory.py
Ben-Cheng: changed it into an exception

topo_processor/cog/execution.py
Ben-Cheng: we have reverted this back. let's open another issue for this

blacha (approved):

PR #163, opened 2021-03-29
fix: non tiff files can also be uploaded

MDavidson17:
Fixes: TDE-29


topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -13,52 +13,57 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
...
             self.read_csv()
 
-        item.source_path_basename = os.path.splitext(os.path.basename(item.source_path))[0]
-        if item.source_path_basename not in self.raw_metadata:
-            raise Exception(f"{item.source_path_basename} cannot be found in the csv.")
-        item_metadata = self.raw_metadata[item.source_path_basename]
+        filename = os.path.splitext(os.path.basename(asset.path))[0]
+
+        if filename not in self.raw_metadata:
+            asset.valid.add_error("Asset not found in CSV file", self.name, "Asset not found in CSV file")
```
blacha, line 26: I think we should make the second text optional.

we only need to record a error if a error is thrown

topo_processor/stac/item_factory.py
```diff
@@ -6,34 +6,30 @@
 from topo_processor.data.data_transformers import data_transformer_repo
 from topo_processor.metadata.metadata_loaders import metadata_loader_repo
 from topo_processor.metadata.metadata_validators import metadata_validator_repo
-from topo_processor.stac import DataType, Item
-from topo_processor.stac.collection_store import get_collection
+from topo_processor.stac.store import get_asset, item_store
 from topo_processor.util import time_in_ms
 
 
-async def create_items(source_dir: str, data_type: DataType, target: str, temp_dir: str) -> None:
```
Ben-Cheng, line 14: should we split it into two function? create_assets / create_items

topo_processor/stac/item_factory.py
```diff
@@ -6,34 +6,30 @@
...
     get_log().debug("Items Created", source_dir=source_dir, duration=time_in_ms() - start_time)
 
 
 async def process_item(item):
-    await metadata_loader_repo.load_metadata(item)
-    if item.is_valid:
+    if item.check_validity:
         await metadata_validator_repo.validate_metadata(item)
-    if item.is_valid:
+    if item.check_validity:
```
Ben-Cheng, line 34: I reckon change this back to is_valid.. more comment on item.py

topo_processor/stac/item.py
```diff
@@ -1,58 +1,50 @@
...
-                file_ext=os.path.splitext(source_path)[1],
-                needs_upload=True,
-            )
-        }
+        self.id = item_id
+        self.stac_extensions = set(["file"])
+        self.collection = None
+        self.assets = []
+
+    def check_validity(self):
```
Ben-Cheng, line 32: I reckon this should happen in metadata_loader_repo.py

topo_processor/metadata/metadata_loaders/metadata_loader_repo.py
```diff
@@ -17,23 +16,21 @@ class MetadataLoaderRepository:
...
-                        await loader.load_metadata(item)
-                    except Exception as error_msg:
-                        item.is_valid = False
-                        item.error_msgs.append(str(error_msg))
-                        get_log().warning(
-                            f"Metadata Load Failed: {error_msg}", loader=loader.name, source_path=item.source_path
-                        )
+                        await loader.load_metadata(asset)
+                        if not asset.is_valid:
+                            break
```
Ben-Cheng, line 27: i think add asset.item.is_valid = False

topo_processor/stac/store.py
```diff
@@ -0,0 +1,28 @@
...
+        asset = Asset(path)
+        asset_store[path] = asset
+    return asset_store[path]
+
+
+def get_item(item_id: str) -> Item:
+    if item_id not in item_store:
+        item = Item(item_id)
+        item_store[item_id] = item
+    return item_store[item_id]
```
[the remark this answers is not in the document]
MDavidson17: I asked Blayne about what the best way to run these functions is/are we tested the following on a dictionary of 100,000:
1. ```def a(title: str): 
    if title not in store:
        collection = title
        store[title] = collection
    return store[title]

2. ```def b(title: str):
    collection = store.get(title)
    if collection is None:
        collection = title
        store[title] = collection
    return collection

3. ```def c(title: str):
    if title not in store:
        collection = title
        store[title] = collection
        return collection
    return store[title]

Surprisingly they all run in basically the same amount of time:

topo_processor/stac/item_factory.py
MDavidson17: I also prefer is_valid, however then the item with have both a function and a variable with the name name which causes errors:
`TypeError: 'bool' object is not callable`

topo_processor/stac/store.py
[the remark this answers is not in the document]
Ben-Cheng: that's awesoe

Ben-Cheng (approved):

topo_processor/metadata/metadata_loaders/metadata_loader_repo.py
MDavidson17: metadata loader imagery historic will return before an item is created so at this point `asset.item = None`.

But you raise a good point we will have to be consistent with this when processing future datasets :thinking:

PR #213, opened 2021-05-04
feat: read from and write to both s3 and local

Ben-Cheng:


Ben-Cheng:
transformer not working on s3 yet

MDavidson17 (approved):

PR #224, opened 2021-05-12
feat: cog from s3

MDavidson17:
uses vsis3 and adds aws credentials as env variables to docker command to take s3 input files for cogging.

### DIFF
--- a/topo_processor/cog/create_cog.py
+++ b/topo_processor/cog/create_cog.py
@@ -1,10 +1,15 @@
 import os
 
 from topo_processor.cog.command import Command
+from topo_processor.util.aws_credentials import get_credentials
 
 
 def create_cog(input_path: str, output_path: str, compression_method: str) -> str:
     cmd = Command("gdal_translate", {"container": "osgeo/gdal", "tag": "ubuntu-small-latest"})
+    if input_path.startswith("/vsis3"):
+        cmd.env(f"AWS_ACCESS_KEY_ID={get_credentials().access_key}")
+        cmd.env(f"AWS_SECRET_ACCESS_KEY={get_credentials().secret_key}")
+        cmd.env(f"AWS_SESSION_TOKEN={get_credentials().token}")
     cmd.mount(input_path)
     cmd.mount(os.path.dirname(output_path))
     cmd.arg(input_path)
--- a/topo_processor/cog/create_cog.py
+++ b/topo_processor/cog/create_cog.py
@@ -6,10 +6,13 @@ from topo_processor.util.aws_credentials import get_credentials
 
 def create_cog(input_path: str, output_path: str, compression_method: str) -> str:
     cmd = Command("gdal_translate", {"container": "osgeo/gdal", "tag": "ubuntu-small-latest"})
-    if input_path.startswith("/vsis3"):
+
+    if input_path.startswith("s3://"):
         cmd.env(f"AWS_ACCESS_KEY_ID={get_credentials().access_key}")
         cmd.env(f"AWS_SECRET_ACCESS_KEY={get_credentials().secret_key}")
         cmd.env(f"AWS_SESSION_TOKEN={get_credentials().token}")
+        input_path = f"/vsis3/{input_path.replace('s3://', '')}"
+
     cmd.mount(input_path)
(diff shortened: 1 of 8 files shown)


topo_processor/data/data_transformers/data_transformer_imagery_historic.py
```diff
@@ -23,18 +23,26 @@ def is_applicable(self, item: Item) -> bool:
...
-            if is_tiff(asset.source_path):
-                start_time = time_in_ms()
-                output_path = os.path.join(item.collection.get_temp_dir(), f"{ulid.ULID()}.tiff")
+            if not is_tiff(asset.source_path):
+                return
+
+            start_time = time_in_ms()
+            output_path = os.path.join(item.collection.get_temp_dir(), f"{ulid.ULID()}.tiff")
+
+            if asset.source_path.startswith("s3://"):
```
blacha, line 32: I think this change should be handled inside the `create_cog`

and we should have a `path.is_s3(source_path)` function somewhere rather than doing startswith `s3://`

topo_processor/cog/create_cog.py
```diff
@@ -1,10 +1,19 @@
...
 from topo_processor.cog.command import Command
+from topo_processor.file_system.get_fs import is_s3_path
+from topo_processor.util.aws_credentials import get_credentials
 
 
 def create_cog(input_path: str, output_path: str, compression_method: str) -> str:
     cmd = Command("gdal_translate", {"container": "osgeo/gdal", "tag": "ubuntu-small-latest"})
+
+    if is_s3_path(input_path):
+        cmd.env(f"AWS_ACCESS_KEY_ID={get_credentials().access_key}")
```
blacha, line 12: credentials may vary based on source/destination, 

I think we need a `get_credentials(path)` which then splits it into the `bucket` and `key` component and loads the credentials based off the `bucket` name failing back to the OS `AWS_PROFILE` if none are configured.

topo_processor/util/aws_credentials.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+import boto3
+
+credentials = None
+
+
+def get_credentials():
+    global credentials
+    if not credentials:
```
blacha, line 10: long running commands will need the credentials refreshed.

topo_processor/util/aws_credentials.py
```diff
@@ -3,11 +3,32 @@
...
-def get_credentials():
+def get_credentials(bucket_name: str):
     global credentials
-    if not credentials:
-        session = boto3.Session(profile_name=os.getenv("AWS_PROFILE"))
-        credentials = session.get_credentials()
+    if credentials:
+        return credentials
+
+    session = boto3.Session(profile_name=os.getenv("AWS_PROFILE"))
```
blacha, line 14: should move this outside of the function, we do not need a new session for every invocation

topo_processor/util/aws_credentials.py
```diff
@@ -3,11 +3,32 @@
...
+role_dict = {}
 
 
-def get_credentials():
+def get_credentials(bucket_name: str):
     global credentials
-    if not credentials:
-        session = boto3.Session(profile_name=os.getenv("AWS_PROFILE"))
-        credentials = session.get_credentials()
+    if credentials:
```
blacha, line 11: this check cannot be done before the `if bucket_name in role_dict` check.

as it will ignore the bucket specific checks.

topo_processor/util/aws_credentials.py
```diff
@@ -3,11 +3,32 @@
...
+
+    session = boto3.Session(profile_name=os.getenv("AWS_PROFILE"))
+
+    if bucket_name in role_dict:
+        if not role_dict[bucket_name]["client"]:
+            client = boto3.client()
+            client.assume_role(RoleArn=role_dict[bucket_name]["role"])
+            role_dict[bucket_name]["client"] = client
+        cred = role_dict[bucket_name]["client"].get("Credentials")
+        credentials = {
```
blacha, line 22: it would be easiest to just return this object here rather than setting it.

blacha (approved):

PR #276, opened 2021-07-05
fix: populate stac metadata hrefs with relative links

MDavidson17:


blacha (approved):

topo_processor/util/transfer_collection.py
```diff
@@ -9,19 +9,34 @@
 
 async def transfer_collection(collection: Collection, target: str):
     stac_collection = collection.create_stac()
+    stac_collection.set_self_href("./collection.json")
```
paulfouquet, line 12: I would recommend setting this in the collection.create_stac() function. What do you think?

paulfouquet (changes requested):
topo_processor/cli/tests/main_test.py
As you've modified this file, could you please remove line 26?

Also, check my comment on transfer_collection.py

paulfouquet (approved):

PR #277, opened 2021-07-06
fix: recursive file scanning

MDavidson17:
use fsspec.walk rather than fsspec.ls

[and add Paul as codeowner and remove Ben C]


topo_processor/stac/item_factory.py
```diff
@@ -21,12 +21,15 @@ async def process_directory(source_dir: str) -> None:
...
-        if is_s3_path(source_dir):
-            asset = get_asset(f"s3://{file_}")
-        else:
-            asset = get_asset(file_)
-        assets_to_process.append(metadata_loader_repo.load_metadata(asset))
+    for (path, _, files) in fs.walk(source_dir):
+        if not files:
+            continue
+        for file_ in files:
+            if is_s3_path(source_dir):
```
blacha, line 28: this is super annoying, imagine if we now had to support azure and google cloud... everywhere we do this we would now need 4 if statements :( is there something in the `fs` that does this for us?

topo_processor/file_system/join_overlapping_paths.py
```diff
@@ -0,0 +1,17 @@
+import difflib
+
+
+def join_overlapping_paths(path_one: str, path_two: str) -> str:
```
blacha, line 4: would be a good idea to have some unit tests for this.

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*   @blacha @MDavidson17 @Ben-Cheng 
+*   @blacha @paulfouquet @MDavidson17
```
blacha, line 1: this should really be a separate pr

blacha (approved):

PR #278, opened 2021-07-07
Fix/update stac v1.0.0

MDavidson17:
Update stac metadate to validate against STAC v1.0.0
nb: some workarounds as pypi pystac is outdated
nb: some metadata, i.e.- extents, are not completely implemented (they contain placeholder data)

### DIFF
--- a/topo_processor/stac/collection.py
+++ b/topo_processor/stac/collection.py
@@ -9,7 +9,7 @@ from linz_logger import get_log
 
 from topo_processor.util import Validity
 
-GLOBAL_PROVIDERS = [pystac.Provider(name="LINZ", description="Land Information New Zealand", roles=["Host"])]
+GLOBAL_PROVIDERS = [pystac.Provider(name="LINZ", description="Land Information New Zealand", roles=["host"])]
 if TYPE_CHECKING:
     from .item import Item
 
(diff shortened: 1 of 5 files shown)


topo_processor/stac/collection.py
```diff
@@ -58,9 +58,9 @@ def delete_temp_dir(self):
     def create_stac(self) -> pystac.Collection:
         stac = pystac.Collection(
             id=str(ulid.ULID()),
-            description=None,
-            license=None,
+            description="STAC Collection Metadata for Historical Imagery",
+            license="MIT",
```
blacha, line 62: We are not MIT ;)

topo_processor/stac/collection.py
```diff
@@ -58,9 +58,9 @@ def delete_temp_dir(self):
     def create_stac(self) -> pystac.Collection:
         stac = pystac.Collection(
             id=str(ulid.ULID()),
-            description=None,
-            license=None,
+            description="STAC Collection Metadata for Historical Imagery",
```
blacha, line 61: I think historical imagery specific stuff should be isolated into historical imagery specific stuff.

blacha (approved):

PR #353, opened 2021-09-21
feat: use camera extension

MDavidson17:
changes:
- camera extension is now added and used in metadata_loader_imagery_historic.py
- camera extension requires integers so convert_value function was added in conversions.py
- stac is validated in metadata_validator_stac.py
- test added


topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -67,6 +71,7 @@ async def load_metadata(self, asset: Asset) -> None:
                 "linz:photo_version": asset_metadata["photo_version"],
             }
         )
+        item.add_extension("https://linz.github.io/stac/v0.0.2/camera/schema.json")
```
blacha, line 74: Did we decide on if `camera_sequence_no` and `nominal_focal_length` are both null do we need a `camera` schema added?

blacha:
The pull request tile is also not conventional, could you please change it :)

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -46,8 +50,8 @@ async def load_metadata(self, asset: Asset) -> None:
                 "linz:photo_no": asset_metadata["photo_no"],
                 "linz:alternate_survey_name": asset_metadata["alternate_survey_name"],
                 "linz:camera": asset_metadata["camera"],
-                "linz:camera_sequence_no": asset_metadata["camera_sequence_no"],
-                "linz:nominal_focal_length": asset_metadata["nominal_focal_length"],
+                "camera:sequence_number": asset_metadata["camera_sequence_no"],
```
blacha, line 53: also possibly super pedantic, but it might be nice to do a separate `update` per extension, maybe a `add_camera_metadata()` or something

topo_processor/util/conversions.py
```diff
@@ -0,0 +1,13 @@
+def convert_value(value):
```
dwsilk, line 1: Let's be more intentional for each key/value that we are handling - this method will cause issues where something looks like an integer but should actually be stored as a string.

topo_processor/metadata/metadata_validators/tests/metadata_validator_stac_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+@pytest.mark.asyncio
+async def test_check_validity():
+    source_path = os.path.join(os.getcwd(), "test_data", "tiffs", "SURVEY_1", "CONTROL.tiff")
+    asset = Asset(source_path)
+    item = Item("item_id")
+    item.add_asset(asset)
+    item.properties.update({"camera:nominal_focal_length": "string"})
+    item.properties.update({"camera:sequence_number": 1234})
+    item.add_extension("https://linz.github.io/stac/v0.0.2/camera/schema.json")
```
dwsilk, line 17: Move extension version numbers to their own file so that we only need to bump version numbers in one location.

dwsilk (changes requested):

topo_processor/metadata/metadata_validators/tests/metadata_validator_stac_test.py
blacha: I am unsure if we can bump all the versions of  the linz extensions at the same time, but we could have a `StacCameraExtension` object which that includes the specific version/uri we are using

dwsilk: Yea we were discussing one version number for each extension, including community/core extensions. Risky having hard coded extension numbers in different parts of the code/tests and having to make sure you've found them all each time one of them needs to be bumped.

dwsilk: And yes URI might be better than version number so that they could be swapped out e.g. when a "LINZ extension" becomes a "community extension".

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
MDavidson17: The camera extension currently doesn't allow for null values, so I've upadated the code to only add the fields and extension if the values exist.

dwsilk (approved):

paulfouquet (approved):
Looks good!

blacha (approved):

blacha (approved):

amfage (approved):

amfage (approved):

PR #402, opened 2021-11-03
feat: implement historical imagery stac extension

MDavidson17:
note:
collection validation will be implemented as part of [TDE-170](https://linzgovt.atlassian.net/jira/software/projects/TDE/boards/607?selectedIssue=TDE-170)
validation of the collection currently fails as provideders are to be added in [TDE-193](https://linzgovt.atlassian.net/jira/software/projects/TDE/boards/607?selectedIssue=TDE-193)

### DIFF
--- a/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
+++ b/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
@@ -78,7 +78,7 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
         item.properties.update(
             {
                 "mission": title,
-                "platform": "fixed-wing aircraft",
+                "platform": "Fixed-wing Aircraft",
                 "instruments": [metadata_row["camera"]],
                 "linz:photo_type": metadata_row["photo_type"],  # to be replaced by Linz:geospatial_type
             }
--- a/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
+++ b/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
@@ -115,8 +115,8 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
             )
         else:
             item.properties.update({"proj:centroid": centroid})
-            if stac.StacExtensions.projection.value not in item.stac_extensions:
-                item.add_extension(stac.StacExtensions.projection.value)
+            if stac.HistoricalStacExtensions.projection.value not in item.stac_extensions:
+                item.add_extension(stac.HistoricalStacExtensions.projection.value)
 
     def add_mission(self, item: Item, asset_metadata: Dict[str, str]):
         survey = asset_metadata.get("survey", None)
@@ -150,7 +150,7 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
         camera_properties["camera:nominal_focal_length"] = string_to_number(asset_metadata["nominal_focal_length"])
 
         item.properties.update(remove_empty_strings(camera_properties))
-        item.add_extension(stac.StacExtensions.camera.value)
+        item.add_extension(stac.HistoricalStacExtensions.camera.value)
 
     def add_film_metadata(self, item: Item, asset_metadata: Dict[str, str]):
         film_properties = {}
@@ -161,7 +161,7 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
         film_properties["film:physical_size"] = asset_metadata["format"]
 
         item.properties.update(remove_empty_strings(film_properties))
-        item.add_extension(stac.StacExtensions.film.value)
+        item.add_extension(stac.HistoricalStacExtensions.film.value)
 
     def add_aerial_photo_metadata(self, item: Item, asset_metadata: Dict[str, str]):
         aerial_photo_properties = {}
@@ -188,7 +188,7 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
             aerial_photo_properties["aerial-photo:scale"] = scale
 
         item.properties.update(remove_empty_strings(aerial_photo_properties))
-        item.add_extension(stac.StacExtensions.aerial_photo.value)
+        item.add_extension(stac.HistoricalStacExtensions.aerial_photo.value)
 
     def add_scanning_metadata(self, item: Item, asset_metadata: Dict[str, str]):
         scanning_properties = {}
@@ -199,7 +199,7 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
             scanning_properties["scan:scanned"] = quarterdate_to_datetime(asset_metadata["when_scanned"])
 
         item.properties.update(remove_empty_strings(scanning_properties))
-        item.add_extension(stac.StacExtensions.scanning.value)
+        item.add_extension(stac.HistoricalStacExtensions.scanning.value)
 
     def add_datetime_property(self, item: Item, asset_metadata: Dict[str, str]):
         item_date = asset_metadata.get("date", None)
--- a/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
+++ b/topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
@@ -67,19 +67,18 @@ class MetadataLoaderImageryHistoric(MetadataLoader):
 
         item.properties.update(
             {
-                "linz:sufi": metadata_row["sufi"],
-                "linz:survey": metadata_row["survey"],
-                "linz:alternate_survey_name": metadata_row["alternate_survey_name"],
-                "linz:camera": metadata_row["camera"],
-                "linz:photocentre_lat": metadata_row["photocentre_lat"],
-                "linz:photocentre_lon": metadata_row["photocentre_lon"],
-                "linz:photo_type": metadata_row["photo_type"],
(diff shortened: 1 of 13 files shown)


topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -109,6 +106,31 @@ def read_csv(self, metadata_file: str = "") -> None:
...
+            "lon": string_to_number(asset_metadata.get("photocentre_lon", None)),
+        }
+        if not all(centroid.values()) or "" in centroid.values():
+            item.add_warning(
+                msg="Skipped Record",
+                cause=self.name,
+                e=Exception(f"stac field 'proj:centroid' has None values: {centroid}"),
+            )
+        else:
+            item.properties.update({"proj:centroid": centroid})
```
blacha, line 121: I think its clearer to do `item.properties['proj:centorid'] = centorid` than doing a `.update` and creating a whole new dict for it.

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -109,6 +106,31 @@ def read_csv(self, metadata_file: str = "") -> None:
 
         self.is_init = True
 
+    def add_centroid(self, item: Item, asset_metadata: Dict[str, str]):
+        centroid = {
+            "lat": string_to_number(asset_metadata.get("photocentre_lat", None)),
+            "lon": string_to_number(asset_metadata.get("photocentre_lon", None)),
+        }
+        if not all(centroid.values()) or "" in centroid.values():
```
blacha, line 114: would be good to validate that these are actual lat/lon and not inverted lon/lat or random numbers?

also is `0` considered a warning?

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -109,6 +106,31 @@ def read_csv(self, metadata_file: str = "") -> None:
...
+            item.properties.update({"proj:centroid": centroid})
+            item.add_extension(stac.StacExtensions.projection.value)
+
+    def add_mission(self, item: Item, asset_metadata: Dict[str, str]):
+        survey = asset_metadata.get("survey", None)
+        if not survey or survey == "0" or survey == "":
+            alternative_survey = asset_metadata.get("alternative_survey", None)
+            if alternative_survey and alternative_survey != "":
+                item.properties.update(remove_empty_strings({"mission": alternative_survey}))
+                return
```
paulfouquet, line 130: You don't need return as there is not further statement

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -109,6 +106,31 @@ def read_csv(self, metadata_file: str = "") -> None:
...
+            )
+        else:
+            item.properties.update({"proj:centroid": centroid})
+            item.add_extension(stac.StacExtensions.projection.value)
+
+    def add_mission(self, item: Item, asset_metadata: Dict[str, str]):
+        survey = asset_metadata.get("survey", None)
+        if not survey or survey == "0" or survey == "":
+            alternative_survey = asset_metadata.get("alternative_survey", None)
+            if alternative_survey and alternative_survey != "":
```
paulfouquet, line 128: In the other case `mission` will never be added. As it is a required field, do we want a warning for that?

topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
```diff
@@ -226,3 +226,43 @@ def test_spatial_metadata_collection_polygon():
...
+    metadata = {
+        "survey": "12345",
+        "alternative_survey": "",
+    }
+    metadata_loader_imagery_historic = MetadataLoaderImageryHistoric()
+    metadata_loader_imagery_historic.add_mission(item, asset_metadata=metadata)
+    assert item.properties["mission"] == "12345"
+
+
+def test_mission_metadata_added_by_alternative_survey():
```
paulfouquet, line 258: I would add a test with `survey = 0` or `""` and `alternative_survey = ""`

paulfouquet (approved):
Looks good, just few questions/suggestions

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
MDavidson17: we could put a warning here, I had left it for the metadata_validator_stac to pick it up?

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
MDavidson17: I added some checks to validate the numbers are within either -90-90 or -180-180. We could make this more specific to NZ (but as is crosses the dateline the nz bbox is not quite accurate)?

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
paulfouquet: I agree with that approach too. However we should clarify this with @dwsilk and apply this (leaving the jsonschema validation to raise the error) to every required field.

paulfouquet (approved):

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -192,3 +198,44 @@ def add_datetime_property(self, item: Item, asset_metadata: Dict[str, str]):
...
+            item.add_extension(stac.StacExtensions.projection.value)
+
+    def is_valid_centroid(self, item: Item, centroid) -> bool:
+        if not all(centroid.values()) or "" in centroid.values():
+            item.add_warning(
+                msg="Skipped Record",
+                cause=self.name,
+                e=Exception(f"stac field 'proj:centroid' has None values: {centroid}"),
+            )
+            return False
```
blacha, line 219: this has a weird mix of elif if and returns, I would stick to one, (I like returns)

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -192,3 +198,44 @@ def add_datetime_property(self, item: Item, asset_metadata: Dict[str, str]):
...
+        if any(isinstance(value, str) for value in centroid.values()):
+            item.add_warning(
+                msg="Skipped Record",
+                cause=self.name,
+                e=Exception(f"stac field 'proj:centroid' has invalid lat value: {centroid['lat']}"),
```
blacha, line 224: This entire function could be simplified to:

```python
if not isinstance(centeroid["lat"], numbers.Number) or centeroid["lat"] > 90 or centeroid["lat"] < -90:
  # lat invalid
  
if not isinstance(centeroid["lon"], numbers.Number) or centeroid["lon"] > 180 or centeroid["lon"] < -180:
  # lon invalid
```

The `numbers.Number` matches both `int` and `float` so will exclude None, str, Dict etc..

It would be great to have a few unit tests for this function.

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -192,3 +206,34 @@ def add_datetime_property(self, item: Item, asset_metadata: Dict[str, str]):
...
+
+    def add_centroid(self, item: Item, asset_metadata: Dict[str, str]):
+
+        centroid = {
+            "lat": string_to_number(asset_metadata.get("photocentre_lat", None)),
+            "lon": string_to_number(asset_metadata.get("photocentre_lon", None)),
+        }
+        if self.is_valid_centroid(item, centroid):
+            item.properties["proj:centroid"] = centroid
+            item.add_extension(stac.StacExtensions.projection.value)
```
amfage, line 218: The projection extension should always be added to the stac_extensions, as the historical imagery extension requires it in stac_extensions (proj:epsg is a required field).

topo_processor/metadata/metadata_loaders/metadata_loader_tiff.py
```diff
@@ -23,18 +25,40 @@ def is_applicable(self, asset: Asset) -> bool:
...
+            crs = tiff.crs.to_epsg()
+        else:
+            crs = None
+        asset.item.properties["proj:epsg"] = crs
+        asset.item.add_extension(stac.StacExtensions.projection.value)
+
+    def add_bands(self, tiff, asset):
+        if ColorInterp.gray in tiff.colorinterp and len(tiff.colorinterp) == 1:
+            asset.properties["eo:bands"] = [{"name": ColorInterp.gray.name, "common_name": "pan"}]
+            asset.item.add_extension(stac.StacExtensions.eo.value)
```
amfage, line 51: The eo extension should always be added to stac_extensions or validation will fail.

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
amfage: I think it is always added when adding proj:epsg

amfage (approved):
Looks good :-)

blacha (approved):

PR #405, opened 2021-11-03
fix(validate): validate STAC Item against item-spec json schema (TDE-204)

paulfouquet:


topo_processor/stac/item.py
```diff
@@ -30,6 +31,7 @@ def __init__(self, item_id: str):
         self.collection = None
         self.geometry_poly = None
         self.assets = []
+        self.schema = 'https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json'
```
blacha, line 34: Is this defined in pystac anywhere? could we just reference it from there?

paulfouquet: I've pushed a change this morning

amfage (approved):

topo_processor/stac/item.py
blacha: Wow that is kinda frustrating how hard it is to get a schema URI from pystac!  maybe we do just want to hard code them :P

blacha (approved):

topo_processor/stac/item.py
paulfouquet: I know right!

PR #410, opened 2021-11-04
chore: Update `jsonschema_rs`

Stranger6667:
### Change Description

I noted, that you use `json.loads` to load JSON Schema [here](https://github.com/linz/topo-processor/blob/master/topo_processor/metadata/metadata_validators/metadata_validator_stac.py#L24), so I implemented `JSONSchema.from_str` to avoid calling Python's `json` at all in such cases. Locally, it gave substantial speedup for large schemas.


paulfouquet (approved):
Thanks for your contribution @Stranger6667

blacha:
Awesome, !hanks @Stranger6667 !

blacha (approved):

PR #413, opened 2021-11-04
ci: build on pull request and pushes

blacha:
Pull requests from external acocunts were not being run through CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
 name: Build
-on: [push]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:


.github/workflows/ci.yml
```diff
@@ -1,5 +1,9 @@
 name: Build
-on: [push]
+on:
```
paulfouquet, line 2: I am not familiar with that. Does that mean the build will run every push on the master branch and also every pull request whatever branch?

paulfouquet (commented):
Question

.github/workflows/ci.yml
blacha: Yes,

pull requests from other peoples accounts are not considered a a "push" so CI will never run, this makes it so if i pull request from say `blacha/topo-processor` the CI system will run.

It will also run on any push to the master branch, this does mean that just random pushes to random branches will not be CI'd until a PR is made.

paulfouquet: Nice!

paulfouquet (approved):

blacha:
ci is failing because of #411

amfage (approved):

PR #415, opened 2021-11-08
fix: asset object key TDE-188

amfage:
[https://linzgovt.atlassian.net/browse/TDE-188](https://linzgovt.atlassian.net/browse/TDE-188)
The Asset Object key was set to the media type.
This is an interim change (until we know more about the assets there will be) to set it to "visual" if the type is geotiff, then to follow the previous logic and set to the file extension if there is no media type. Then to the checksum if there is no extension.

### DIFF
--- a/topo_processor/stac/asset.py
+++ b/topo_processor/stac/asset.py
@@ -41,6 +41,16 @@ class Asset(Validity):
             self.properties["file:checksum"] = multihash_as_hex(self.source_path)
         return self.properties["file:checksum"]
 
+    def get_key(self) -> str:
+        if self.content_type:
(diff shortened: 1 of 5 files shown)


topo_processor/stac/asset.py
```diff
@@ -41,6 +41,16 @@ def get_checksum(self) -> str:
             self.properties["file:checksum"] = multihash_as_hex(self.source_path)
         return self.properties["file:checksum"]
 
+    def get_key(self) -> str:
+        if self.content_type:
```
paulfouquet, line 45: You might not need to test if `content_type` is not `None`

paulfouquet (approved):
Looks good to me, just a suggestion

amfage:
Change of approach needed.

PR #434, opened 2021-11-12
fix: asset object key redo (TDE-188)

amfage:
[https://linzgovt.atlassian.net/browse/TDE-188](https://linzgovt.atlassian.net/browse/TDE-188)

Note: should I check against the enumerator when adding the key in `topo_processor/util/transfer_collection.py`, rather than assume an enumerator value has been used?

### DIFF
--- a/topo_processor/stac/asset_key.py
+++ b/topo_processor/stac/asset_key.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class AssetKey(Enum):
+    Visual = "visual"
+    Thumbnail = "thumbnail"
--- a/topo_processor/stac/asset_key.py
+++ b/topo_processor/stac/asset_key.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
 
-class AssetKey(Enum):
+class AssetKey(str, Enum):
     Visual = "visual"
     Thumbnail = "thumbnail"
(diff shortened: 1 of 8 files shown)


topo_processor/util/tests/transfer_collection_test.py
```diff
@@ -25,19 +26,40 @@ def test_fail_on_duplicate_assets(setup):
...
-    collection.license = "face_license"
+    collection.license = "fake_license"
     item = Item("item_id")
     item.datetime = datetime.now()
     collection.add_item(item)
     item.collection = collection
 
     cog_1 = Asset("./test_data/tiffs/SURVEY_1/CONTROL.tiff")
     cog_1.target = "fake_title/fake_target.tiff"
+    cog_1.key_name = AssetKey.Visual.value
```
paulfouquet, line 37: Instead of using `enum.value` you can define AssetKey(str, Enum), so the `enum` will return the `string` value. i.e. `AssetKey.Visual == "visual" # True`

topo_processor/stac/asset_key.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class AssetKey(Enum):
```
paulfouquet, line 4: `class AssetKey(str, Enum):`
See the [doc](https://docs.python.org/3/library/enum.html#others)

paulfouquet (approved):
I've made a suggestion, but that already looks good to me

paulfouquet (approved):

MDavidson17 (approved):
:+1: 
We should probably also implement this enum fix for our other enum classes at some point.

PR #443, opened 2021-11-18
feat(validate): validation against local schema for tests (TDE-234)

paulfouquet:
### Change Description:
The JSON schema loader can now load a local schema. 
This is for tests only.

### Notes for Testing:
A new test has been added as an example

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,8 @@ pystac = "^1.2.0"
 jsonschema = "^4.2.1"
 jsonschema-rs = "^0.13.1"
 Shapely = "^1.8.0"
+requests = "^2.26.0"
+aiohttp = "^3.8.1"
 
 [tool.poetry.dev-dependencies]
 black = "^21.10b0"
--- a/topo_processor/metadata/metadata_validators/metadata_validator_stac.py
+++ b/topo_processor/metadata/metadata_validators/metadata_validator_stac.py
@@ -19,9 +19,14 @@ class MetadataValidatorStac(MetadataValidator):
 
     def get_validator_from_uri(self, schema_uri: str) -> Any:
         if schema_uri not in self.validator_cache:
-            response = urllib.request.urlopen(schema_uri)
-            schema = response.read()
-            self.validator_cache[schema_uri] = jsonschema_rs.JSONSchema.from_str(schema.decode("utf8"))
+            if schema_uri.startswith("http"):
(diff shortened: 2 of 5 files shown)


topo_processor/metadata/metadata_validators/metadata_validator_stac.py
```diff
@@ -19,9 +19,14 @@ class MetadataValidatorStac(MetadataValidator):
 
     def get_validator_from_uri(self, schema_uri: str) -> Any:
         if schema_uri not in self.validator_cache:
-            response = urllib.request.urlopen(schema_uri)
-            schema = response.read()
-            self.validator_cache[schema_uri] = jsonschema_rs.JSONSchema.from_str(schema.decode("utf8"))
+            if schema_uri.startswith("http"):
```
blacha, line 22: Maybe future work?  This use that fsspec lib? would then let you do `s3://` `file://` and `http://`?

topo_processor/metadata/metadata_validators/metadata_validator_stac.py
```diff
@@ -19,9 +19,14 @@ class MetadataValidatorStac(MetadataValidator):
...
+                response = urllib.request.urlopen(schema_uri)
+                schema = response.read()
+                self.validator_cache[schema_uri] = jsonschema_rs.JSONSchema.from_str(schema.decode("utf8"))
+            else:
+                with open(schema_uri) as file:
+                    schema = json.load(file)
```
blacha, line 28: could load this directly from the file, as you are doing 

Buffer -> json -> rust py to json -> schema

where as you could do

Buffer -> rust buffer to json -> schema

blacha (approved):

amfage (approved):

pyproject.toml
```diff
@@ -17,6 +17,8 @@ pystac = "^1.2.0"
 jsonschema = "^4.2.1"
 jsonschema-rs = "^0.13.1"
 Shapely = "^1.8.0"
+requests = "^2.26.0"
```
blacha, line 20: How come we need these two libs?

paulfouquet: It crashed an error saying that they are required for fsspec

blacha: ahh ok! so they are optional deps for fsspec

blacha (approved):

PR #448, opened 2021-11-23
feat(validate): support LDS cache of Historical Imagery metadata as source (TDE-202)

paulfouquet:
`validate` command now uses the `LDS Cache` by default. **It is currently only implemented for the `Historical Imagery` layer**. By default validate against the latest version of the layer. An optional argument allows the user to validate against a specific metadata file.


topo_processor/stac/lds_cache.py
```diff
@@ -0,0 +1,58 @@
...
+
+class LdsCache:
+    bucket: str
+    role: str
+
+    def __init__(self, bucket_name: str, role_arn: str):
+        self.bucket = bucket_name
+        self.role = role_arn
+
+    def get_layer(self, layer: str, version_number: str = None) -> str:
```
paulfouquet, line 18: Do we want to return the metadata csv as a `buffer`? Currently returning the path to the local temporary metadata file (which is deleted at the end of the process). I've made the system downloading the file as it's supposed to be `gzip` so I could not load the file content as I do for the `collection.json`

topo_processor/stac/lds_cache.py
```diff
@@ -0,0 +1,58 @@
...
+
+        metadata_file_name = self.get_metadata_file_name(layer, version_number)
+        metadata_file_path = lds_cache_local_tmp_folder + "/" + metadata_file_name
+        s3_download(
+            build_s3_path(self.bucket, layer + "/" + metadata_file_name),
+            metadata_file_path,
+            self.role,
+        )
+
+        if os.path.isfile(metadata_file_path):
```
paulfouquet, line 33: What is the best way to handle this? If the `s3_download` run properly (or at least `boto3` not raising any exception), could we have an issue that the file does not exist somehow?

topo_processor/stac/lds_cache.py
```diff
@@ -0,0 +1,58 @@
...
+            if is_gzip_file(metadata_file_path):
+                decompress_file(metadata_file_path)
+            return metadata_file_path
+        else:
+            raise Exception(f"{metadata_file_path} not found")
+
+    def get_collection(self, layer: str) -> pystac.Collection:
+        return pystac.Collection.from_dict(load_file_content(self.bucket, layer + "/collection.json", role_arn=self.role))
+
+    @staticmethod
```
paulfouquet, line 43: Should these `static methods` be implemented in another `py` file?

topo_processor/util/aws_credentials.py
```diff
@@ -27,3 +31,18 @@ def get_credentials(bucket_name: str):
         default_credentials["secret_key"] = session_credentials.secret_key
         default_credentials["token"] = session_credentials.token
     return default_credentials
+
+
+def assumed_role_session(role_arn: str, base_session: botocore.session.Session = None) -> boto3.Session:
```
paulfouquet, line 36: This method is temporary. We need to discuss how we want to handle the authentication/role assumption in this project.

topo_processor/util/configuration.py
```diff
@@ -0,0 +1,7 @@
+from dotenv import dotenv_values
+
+configuration = dotenv_values(".env")
```
paulfouquet, line 3: Could also load that as part as the `Environment variables`

topo_processor/util/files.py
```diff
@@ -0,0 +1,37 @@
...
+
+def is_csv(path: str) -> bool:
+    return path.endswith(".csv")
+
+
+def is_gzip_file(file_path: str) -> bool:
+    with open(file_path, "rb") as file:
+        # gzip magic number == "1f 8b"
+        return file.read(2) == b"\x1f\x8b"
+        # FIXME: check Header
```
paulfouquet, line 15: Something better has to be done

topo_processor/util/tests/configuration_test.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+from topo_processor.util.configuration import configuration
+
+
+def test_configuration():
+    assert configuration["LDS_CACHE_LOCAL_TMP_FOLDER"] == "temp"
```
paulfouquet, line 7: Might remove this test as I'm basically testing `dotenv` library...

paulfouquet (commented):
Made some comments on my draft

topo_processor/cli/validate.py
```diff
@@ -3,54 +3,73 @@
...
-from topo_processor.file_system.get_fs import is_s3_path
+from topo_processor.stac.data_type import DataType
 from topo_processor.stac.validation import validate_stac
 from topo_processor.util import time_in_ms
+from topo_processor.util.configuration import lds_cache_local_tmp_folder
+from topo_processor.util.files import empty_dir
 
 
 @click.command()
+@click.option(
```
paulfouquet, line 14: Will need to update the `README.md`

topo_processor/util/tests/aws_files_test.py
```diff
@@ -0,0 +1,25 @@
...
+    This function creates a temporary directory and deletes it after each test.
+    See following link for details:
+    https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended
+    """
+    target = mkdtemp()
+    yield target
+    shutil.rmtree(target)
+
+
+# Add test with AWS mock
```
paulfouquet, line 21: I'm not sure how far we want to test those methods which call AWS via `boto3`. There is `botocore.stubber` and `moto` library that could do the trick.

topo_processor/util/configuration.py
```diff
@@ -0,0 +1,7 @@
...
+
+lds_cache_bucket = configuration["LDS_CACHE_BUCKET"]
+lds_cache_read_role = configuration["LDS_CACHE_READ_ROLE"]
+lds_cache_local_tmp_folder = configuration["LDS_CACHE_LOCAL_TMP_FOLDER"]
```
paulfouquet, line 7: Could also generate a temporary folder at each `validate`/`upload` run, passing it into arguments and delete it at the end

topo_processor/cli/validate.py
```diff
@@ -3,54 +3,73 @@
...
+    "-l",
+    "--layer",
+    required=False,
+    help="The layer number to validate. Take the last version by default.",
```
blacha, line 18: For this to work outside historical imagery. the layerId will have to be hard coded into the historical imagery loader.

topo_processor/cli/validate.py
```diff
@@ -3,54 +3,73 @@
...
+)
+@click.option(
+    "-v",
+    "--version",
+    required=False,
+    help="(OPTIONAL) The specific version number to validate.",
+)
 @click.option(
```
blacha, line 26: we should remove the source parameter from here too. 

We could override the source location possibly with a environment variable if we really want to run custom locations ideally we should just have a 

```
--metadata ./foo.csv
--metadata s3://foo/bar.csv
```

topo_processor/util/aws_files.py
```diff
@@ -0,0 +1,37 @@
...
+    url_o = urlparse(source_path)
+    bucket_name = url_o.netloc
+    object_name = url_o.path[1:]
+
+    session = assumed_role_session(role_arn)
+    s3 = session.resource("s3")
+    try:
+        s3.Bucket(bucket_name).download_file(object_name, dest_path)
+    except Exception as e:
+        get_log().error(f"An error has occurred while downloading {source_path}: {e}")
```
blacha, line 21: should log these as key value pairs

topo_processor/util/aws_files.py
```diff
@@ -0,0 +1,37 @@
...
+
+def load_file_content(bucket_name: str, object_path: str, role_arn: str):
+    session: boto3.Session = assumed_role_session(role_arn)
+    s3 = session.client("s3")
+    object_content = s3.get_object(Bucket=bucket_name, Key=object_path)
+
+    if object_path.endswith(".json"):
+        return json.loads(object_content["Body"].read().decode("utf-8"))
+
+    return object_content["Body"].read().decode("utf-8")
```
blacha, line 33: I dont think you can assume all of these will be utf8?

topo_processor/util/configuration.py
blacha: temporary files should go into `tempfile.TemporaryFile()`

topo_processor/util/files.py
blacha: the s3 response will include the `Content-Encoding` header which will tell you if the file is gzip encoded. 

otherwise this is the only way to determine if a file on disk is actually gzipd

topo_processor/util/files.py
```diff
@@ -0,0 +1,37 @@
...
+        os.remove(os.path.join(dir, f))
+
+
+def decompress_file(file_path: str) -> None:
+    input: gzip.GzipFile = None
+
+    try:
+        input = gzip.GzipFile(file_path, "rb")
+    except gzip.BadGzipFile as e:
+        get_log().error(f"An error occurred while trying to decompress {file_path}: {e.strerror}")
```
blacha, line 29: use key value paris for logs

topo_processor/cli/validate.py
paulfouquet: OK, always the latest version by default then

topo_processor/util/aws_files.py
paulfouquet: `get_log().error("An error has occurred during the download.", file=source_path, error=e)` ?

topo_processor/util/files.py
paulfouquet: I don't know if testing against `Content-Encoding` from s3 makes sense with the current code as the file is always downloaded?

topo_processor/cli/validate.py
paulfouquet: I'm actually not sure to understand. If we want to run `validate` against another layer than the `Historical Imagery` one, how do you do that?

topo_processor/cli/validate.py
blacha: Yeah, lets make it super simple to run

topo_processor/util/files.py
blacha: I was more thinking if you download from s3, you can check the content-encoding and then gunzip it, otherwise its is unlikely that a local file will be gzip'd

topo_processor/cli/validate.py
blacha: Historical imagery only has one layer that can be used to actually validate it. no other layer has the same table layout that we are expecting.

in the future we will need two layers to actually validate it too.

paulfouquet: So at the moment `validate` will by default and only run against `Historical Imagery`?

topo_processor/util/configuration.py
```diff
@@ -0,0 +1,10 @@
+import os
+from tempfile import mkdtemp
+
+from dotenv import dotenv_values
+
+configuration = dotenv_values(".env")
+
+lds_cache_bucket: str = configuration["LDS_CACHE_BUCKET"]
+aws_role_config_path = os.path.expanduser(configuration["AWS_ROLES_CONFIG"])
+temp_folder: str = mkdtemp()
```
paulfouquet, line 10: This will be instanced on runtime but I don't know how to handle it without having to pass it as argument is all the call stack from `validate`

topo_processor/util/tests/aws_files_test.py
paulfouquet: I haven't implemented that but have created a [JIRA Task](https://linzgovt.atlassian.net/browse/TDE-236).

topo_processor/util/aws_files.py
```diff
@@ -0,0 +1,60 @@
...
+    object_content = s3.Object(bucket_name=bucket_name, key=object_path)
+
+    if object_path.endswith(".json"):
+        return json.loads(object_content.get()["Body"].read())
+
+    return object_content.get()["Body"].read()
+
+
+def build_s3_path(bucket_name: str, object_path: str) -> str:
+    return "s3://" + bucket_name + "/" + (object_path[1:] if object_path.startswith("/") else object_path)
```
amfage, line 60: NIT: Would f-string be nice to use here?

topo_processor/util/gzip.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+
+from linz_logger.logger import get_log
+
+
+def is_gzip_file(file_path: str) -> bool:
+    with open(file_path, "rb") as file:
+        # gzip magic number == "1f 8b"
+        return file.read(2) == b"\x1f\x8b"
+        # FIXME: check Header
```
amfage, line 11: Maybe the try/except is enough? If it doesn't gzip, it's not gzipped I guess :shrug:

topo_processor/util/aws_files.py
blacha: Im suprised boto3 doesnt have a method that does this for us! other option would be a path join 

maybe something like `f"s3://{os.path.join(bucket_name, object_path)}"` ?

topo_processor/util/gzip.py
blacha: I think this is fine for checking files on disk to see if they are `gzip`

topo_processor/util/aws_files.py
blacha: `os.path.join` may have implications under windows (using the wrong slashes) is there a `url.join`?

blacha (approved):

topo_processor/util/gzip.py
paulfouquet: @amfage That would be okay if the `gzip` lib would return a specific `exception` when it's not a `gzipped` file, but it does not. Dealing with a generic `BadGzipFile` with a specific message for each errors (i.e. `BadGzipFile('Not a gzipped file (%r)' % magic)`) is not ideal as our code would have to parse the `error.message` to determine what if the issue. I don't like this approach because it has to rely on a `string` which potentially can change with a next release so you have to maintain it.

amfage (approved):

topo_processor/util/aws_files.py
paulfouquet: There is a [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) but I was not able to get a good output for `s3://` URLs.

PR #457, opened 2021-11-29
refactoring(metadata-loader): merge load_all_metadata with load_metadata (TDE-214)

paulfouquet:
### Change Description:
`load_all_metadata` method does not longer exist and is replace by `load_metadata` with specific arguments.


amfage (approved):
LGTM

PR #501, opened 2022-01-13
ci: add mypy run to the CI workflow (TDE-184)

paulfouquet:
### Change Description:
Mypy configuration is not checking all the current errors by default. That means `mypy .` run on the current code base will not raise any error.
Mypy run is added to the CI workflow. `mypy .` will be run at every build on the GitHub repository.

### Notes for Testing:
Run `mypy .` at the project root. Make sure that it returns `Success: no issues found in [...]`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,6 @@ jobs:
       - name: Test
         run: |
           poetry run pytest --slow --cov topo_processor
+      - name: Mypy
+        run: |
+          poetry run mypy .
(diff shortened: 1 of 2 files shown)


blacha:
This should be a `ci` change not a `feat` as this is not adding any sort of feature to the codebase and is only modifying the `ci` pipeline.

Also scopes are generally for areas of your codebase, so its somewhat weird to have a scope of `(mypy)`, I dont think we have any discussed or defined "scopes" inside of this repo.

otherwise LGTM

amfage (approved):
LGTM :+1:

PR #502, opened 2022-01-13
fix: add missing typing for function arguments [arg-type] (TDE-263)

paulfouquet:
### Change Description:
Add the missing typing for function arguments. That is check by `Mypy` with error code `arg-type`

### Notes for Testing:
run `mypy .` should not display any error.

### DIFF
--- a/topo_processor/cli/validate.py
+++ b/topo_processor/cli/validate.py
@@ -34,7 +34,7 @@ from topo_processor.util.configuration import temp_folder
     is_flag=True,
     help="Use verbose to display trace logs (it might be slower).",
 )
-def main(item: str, collection: str, metadata: str, verbose: str) -> None:
+def main(item: bool, collection: bool, metadata: str, verbose: str) -> None:
     if verbose:
         set_level(LogLevel.trace)
     else:
--- a/topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
+++ b/topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
@@ -292,7 +292,7 @@ def test_invalid_centroid_string() -> None:
     centroid = {"lat": "-41.28509", "lon": "174.77442"}
     metadata_loader_imagery_historic = MetadataLoaderImageryHistoric()
     assert not metadata_loader_imagery_historic.is_valid_centroid(item, centroid)  # type: ignore [arg-type]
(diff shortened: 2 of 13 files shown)


amfage (approved):
:+1:

MDavidson17 (approved):

topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
```diff
@@ -291,7 +291,8 @@ def test_invalid_centroid_string() -> None:
     item = stac.Item(source_path)
     centroid = {"lat": "-41.28509", "lon": "174.77442"}
     metadata_loader_imagery_historic = MetadataLoaderImageryHistoric()
-    assert not metadata_loader_imagery_historic.is_valid_centroid(item, centroid)
+    assert not metadata_loader_imagery_historic.is_valid_centroid(item, centroid)  # type: ignore
```
l0b0, line 294: If this is causing only one type of issue you can be more specific in the comment, such as `# type: ignore[foo]` - "foo" being the name of the test printed by `mypy`. That way other mypy checks still run on the code.

topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
```diff
@@ -291,7 +291,8 @@ def test_invalid_centroid_string() -> None:
...
+    # Ignore typing above because it interfers with our test.
```
amfage, line 295: Tiny NIT: "interferes"

topo_processor/cli/validate.py
```diff
@@ -34,7 +34,7 @@
     is_flag=True,
     help="Use verbose to display trace logs (it might be slower).",
 )
-def main(item: str, collection: str, metadata: str, verbose: str) -> None:
+def main(item: bool, collection: bool, metadata: str, verbose: str) -> None:
```
blacha, line 37: if these are `bool` what exactly does `if item == collection` mean on line 49?

paulfouquet: There might be a prettier way to do that.
That means we call `validate_stac(metadata)` - either with `Item` and `Collection` validation if the user hasn't specify any of them or had specified `True` for both of them.
If one of the flag `Collection` or `Item` has been specified, we call `validate_stac` with those flags regardless which one is `True` or `False`

PR #503, opened 2022-01-13
feat: add gdal_translate options (TDE-80) (TDE-260)

amfage:
Sets docker container environment variable GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR for GDAL COG creation.
[https://linzgovt.atlassian.net/browse/TDE-260](https://linzgovt.atlassian.net/browse/TDE-260)

Sets the following options:
OVERVIEW_COMPRESS=JPEG
OVERVIEW_QUALITY=90
[https://linzgovt.atlassian.net/browse/TDE-80](https://linzgovt.atlassian.net/browse/TDE-80)

### DIFF
--- a/topo_processor/cog/create_cog.py
+++ b/topo_processor/cog/create_cog.py
@@ -27,5 +27,6 @@ def create_cog(input_path: str, output_path: str, compression_method: str, overv
     cmd.arg("-co", "BLOCKSIZE=512")
     cmd.arg("-co", "OVERVIEW_QUALITY=90")
     cmd.arg("-co", "SPARSE_OK=TRUE")
+    cmd.arg("-co", "GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR")
     cmd.arg(output_path)
     return cmd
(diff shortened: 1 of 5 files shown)


dwsilk (approved):

topo_processor/cog/create_cog.py
```diff
@@ -5,27 +5,27 @@
 from topo_processor.util.aws_credentials import Credentials, get_credentials
 
 
-def create_cog(input_path: str, output_path: str, compression_method: str) -> str:
+def create_cog(input_path: str, output_path: str, compression_method: str, overview_compression_method: str) -> str:
```
paulfouquet, line 8: Is there a reason to have an argument for this specific `gdal` parameter? i.e. are we anticipating something?

paulfouquet (commented):
Just a quick question to avoid adding complexity if it's not necessary.

topo_processor/cog/create_cog.py
paulfouquet: IMO we generally should make this kind of change when it actually happens.

amfage: Discussion with @dwsilk @paulfouquet - will hard code both `compression` and `overview_compression`.

paulfouquet (approved):
Looks good!

PR #505, opened 2022-01-16
fix: add missing typing for type arguments [type-arg] (TDE-266)

paulfouquet:
### Change Description:
Add the missing typing for type arguments. That is check by Mypy with error code type-arg

### Notes for Testing:
run `mypy .` should not display any error.

### DIFF
--- a/topo_processor/stac/collection.py
+++ b/topo_processor/stac/collection.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import os
-from datetime import datetime
+from datetime import date, datetime
 from shutil import rmtree
 from tempfile import mkdtemp
-from typing import TYPE_CHECKING, Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List, Set
 
 import pystac
 import ulid
@@ -40,7 +40,7 @@ class Collection(Validity):
     extra_fields: Dict[str, Any]
     linz_geospatial_type: str
 
-    stac_extensions: set
+    stac_extensions: Set[str]
     summaries: Summaries
 
     def __init__(self, title: str):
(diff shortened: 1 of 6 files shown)


amfage (approved):

MDavidson17 (changes requested):
Just a couple of unused imports

topo_processor/util/conversions.py
```diff
@@ -1,6 +1,6 @@
 import re
 from datetime import datetime
-from typing import Dict
+from typing import Any, Dict
```
MDavidson17, line 3: unused `Any' import?

topo_processor/util/aws_credentials.py
```diff
@@ -1,6 +1,6 @@
 import json
 import os
-from typing import Dict
+from typing import Any, Dict
```
MDavidson17, line 3: Unused `Any` here?

topo_processor/stac/collection.py
```diff
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import os
-from datetime import datetime
+from datetime import date, datetime
```
MDavidson17, line 4: unused `Date' import?

topo_processor/util/conversions.py
paulfouquet: Good spot. Fixed in https://github.com/linz/topo-processor/pull/512

topo_processor/util/aws_credentials.py
paulfouquet: Fixed in https://github.com/linz/topo-processor/pull/512

MDavidson17 (approved):
:+1:

PR #507, opened 2022-01-17
fix: add missing typing for [no-any-return] (TDE-267) 

paulfouquet:
### Change Description:
Fix where return a value with an Any type in a function that is annotated to return a non-Any value.


amfage (approved):

topo_processor/stac/asset.py
```diff
@@ -41,7 +41,7 @@ def get_content_type(self) -> str:
             return self.content_type
         return MimeTypes().guess_type(self.target if self.target else self.source_path)[0]
 
-    def get_checksum(self) -> str:
+    def get_checksum(self) -> Any:
```
MDavidson17, line 44: So, there seems to be three options we could pick from and mypy passes?
Is there a rule or best practise for how we should choose?
1. `-> Any`
2. `-> Union[str, Any]`
3. `-> str...
     ... return cast(str, value)`

MDavidson17 (approved):
Just the one comment which is more a question, otherwise looks good to me.

topo_processor/stac/asset.py
paulfouquet: Good spot. It should be `Union[str, Any]` as that is what returns `multihash_as_hex`

topo_processor/util/checksum.py
```diff
@@ -7,7 +8,7 @@
 CHUNK_SIZE = 1024 * 1024  # 1MB
 
 
-def multihash_as_hex(path: str) -> str:
```
l0b0, line 10: `Union[str, Any]` === `Any`

PR #510, opened 2022-01-18
fix: mypy [return-value] (TDE-265)

amfage:
**Change Description:**
Correct return value errors, mypy error code `return-value`
[https://linzgovt.atlassian.net/browse/TDE-265](https://linzgovt.atlassian.net/browse/TDE-265)

**Notes for Testing:**
run `mypy` . should not display any error.

### DIFF
--- a/topo_processor/data/data_transformers/data_transformer_repo.py
+++ b/topo_processor/data/data_transformers/data_transformer_repo.py
@@ -27,7 +27,6 @@ class DataTransformerRepository:
                 except Exception as e:
                     item.add_error(str(e), transformer.name, e)
                     get_log().warning(f"Data Transform Failed: {e}", transformers=transformer.name)
-                    return False
                 get_log().debug(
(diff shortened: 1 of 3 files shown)


paulfouquet (approved):
looks good

MDavidson17 (approved):

topo_processor/data/data_transformers/data_transformer_repo.py
```diff
@@ -27,7 +27,6 @@ def transform_data(self, item: Item) -> None:
                 except Exception as e:
                     item.add_error(str(e), transformer.name, e)
                     get_log().warning(f"Data Transform Failed: {e}", transformers=transformer.name)
-                    return False
```
l0b0, line 30: This changes the semantics of the method, by running the rest of the method rather than returning early when encountering this exception. Are you sure you want to do that?

PR #511, opened 2022-01-18
fix: correct imports to comply with [attr-defined] (TDE-269)

paulfouquet:
### Change Description:
correct import to comply with `[attr-defined]` 

### Notes for Testing:
run `mypy .`


amfage (approved):

amfage (approved):
There are still some imports that are not explicit, but mypy doesn't seem to care about them.

topo_processor/file_system/transfer.py
```diff
@@ -1,11 +1,13 @@
...
 from linz_logger import get_log
 
-from topo_processor.util import time_in_ms
+from topo_processor.util.time import time_in_ms
 
 from .get_fs import get_fs
 
 
-def transfer_file(source_file: str, checksum: str, content_type: str, target_file: str) -> None:
+def transfer_file(source_file: str, checksum: str, content_type: Union[str, None], target_file: str) -> None:
```
MDavidson17, line 10: There seems to be two options for when variables could be none:
- `Union[str, None]`
- `Optional[str]`

Which one is best? (should we be consistent because I've used `Optional[str]`?)

MDavidson17 (approved):
:+1:  just another question but looks good to me

topo_processor/file_system/transfer.py
paulfouquet: It does the same thing (https://docs.python.org/3/library/typing.html#typing.Optional). I vote for `Optional`.

paulfouquet:
> +1 just another question but looks good to me
> 
> Sorry just refreshed the page and saw you marked it as a draft grimacing

Yes, I need to fix an issue following the merging.

topo_processor/file_system/transfer.py
blacha: Also pose the question of `A | B` or `Union[A, B]` 

`A | B` looks better to me as its what typescript uses :)

paulfouquet: I think it is not on the current version we use though

amfage: That's actually the reason I asked about Python versions earlier this week :)

paulfouquet: If we plan to move to `3.10` one day, should we leave this change until then?

blacha: ahh ok! didn't realise it was that new...

MDavidson17 (approved):

PR #512, opened 2022-01-18
fix: remove use of Any in an Union (TDE-278)

paulfouquet:
### Change Description:
Remove the use of `Any` type inside a `Union` to type a functions return.

### DIFF
--- a/topo_processor/util/checksum.py
+++ b/topo_processor/util/checksum.py
@@ -8,9 +8,10 @@ from topo_processor.file_system.get_fs import get_fs
 CHUNK_SIZE = 1024 * 1024  # 1MB
 
 
-def multihash_as_hex(path: str) -> Union[str, Any]:
+def multihash_as_hex(path: str) -> str:
     file_hash = hashlib.sha256()
     with get_fs(path).open(path, "rb") as file:
         while chunk := file.read(CHUNK_SIZE):
             file_hash.update(chunk)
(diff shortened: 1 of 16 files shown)


topo_processor/util/checksum.py
```diff
@@ -8,9 +8,10 @@
...
 
 
-def multihash_as_hex(path: str) -> Union[str, Any]:
+def multihash_as_hex(path: str) -> str:
     file_hash = hashlib.sha256()
     with get_fs(path).open(path, "rb") as file:
         while chunk := file.read(CHUNK_SIZE):
             file_hash.update(chunk)
-    return multihash.to_hex_string(multihash.encode(file_hash.digest(), "sha2-256"))
+    result: str = multihash.to_hex_string(multihash.encode(file_hash.digest(), "sha2-256"))
```
paulfouquet, line 16: `Mypy` was complaining because `to_hex_string` method was not typed as returning a `str`, even if it does return the result of a function typed as returning a `str` (`to_hex_string` returns `hexlify(multihash).decode()` and `decode()` is typed as a `str` return value annotation).
By specified the type for the return value (because we know that `to_hex_string()` always returns a `str`, we force mypy to consider `result` as a `str`.

amfage (approved):

MDavidson17 (approved):

PR #524, opened 2022-01-25
fix: type annotate variable when type is unclear (TDE-284)

paulfouquet:
### Change Description:
Add annotation if variable type is unclear `[var-annotated]`


MDavidson17 (approved):
:+1:

topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
```diff
@@ -196,7 +198,7 @@ def test_add_datetime_property_not_date() -> None:
 def test_spatial_metadata_empty() -> None:
     source_path = "test_abc.tiff"
     item = Item(source_path)
-    metadata = {}
+    metadata: Dict[str, str] = {}
```
l0b0, line 201: This is a single-use variable, so it could be inlined to avoid having to specify a type hint. That would also avoid "drift" from the actual metadata structure, which I think would be `Dict[str, Any]`?

l0b0 (approved):

topo_processor/metadata/metadata_loaders/tests/metadata_loader_imagery_historic_test.py
paulfouquet: I had to type as mypy was complaining about it to be unclear. 
Reading the csv file produces a `dict[str, str]`
```python
with open(csv_path, "r") as csv_text:
  reader = csv.DictReader(csv_text, delimiter=",")
  for row in reader:
```
-> `row: dict[str, str]`

blacha (approved):

PR #525, opened 2022-01-25
fix: typing annotation None to Generator for pytest setup functions (TDE-285)

paulfouquet:
### Change Description:
Use `Generator` type for return type annotation of `pytest` `setup()` functions.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,7 @@
+import shutil
+from tempfile import mkdtemp
+from typing import Generator
+
 import pytest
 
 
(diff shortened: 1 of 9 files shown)


l0b0 (approved):
Looks good, but why not put the duplicate `setup` functions in `conftest.py`?

paulfouquet:
> Looks good, but why not put the duplicate `setup` functions in `conftest.py`?

Good idea, we haven't think about it. I've made a new commit.

l0b0 (approved):

blacha (approved):

MDavidson17 (approved):

PR #526, opened 2022-01-25
fix: centroid has to be an int or a float (TDE-286)

paulfouquet:
### Change Description:
`is_valid_centroid` checks if `centroid` is an `int` or a `float`. That also avoid `mypy` complaining about `Unsupported operand types for < ("int" and "Number")`


l0b0 (approved):

blacha:
A centeroid doesn't have to be a float.. `-170, 40` is a valid centeroid..

Making the logic of the code invalid to fix the type checker is not a good option.

paulfouquet:
> A centeroid doesn't have to be a float.. `-170, 40` is a valid centeroid..
> 
> Making the logic of the code invalid to fix the type checker is not a good option.

As `string_to_number` returns a `Union[float, int, str]`, what we could do is 
```python
if (not isinstance(centroid["lat"], float) or not isinstance(centroid["lat"], int) or centroid["lat"] > 90or centroid["lat"] < -90):
```

Otherwise, ignore this error code.

blacha (approved):

MDavidson17 (approved):

PR #548, opened 2022-02-04
fix: remove metadata validator HI

MDavidson17:
Reasons for deleting this validator:
1. this code currently expects survey -> survey matches but reality is comparing film -> survey.
2. A logic of comparing film -> survey will be used to find images to process (orchestration) so it is redundant here and not worth fixing.


paulfouquet (approved):
:+1:

blacha (approved):

PR #568, opened 2022-02-11
feat: log loading of credentials from ssm and assuming roles

blacha:


topo_processor/util/aws_credentials.py
```diff
@@ -36,11 +37,13 @@ def init_roles() -> None:
     if aws_profile is None:
         return
 
+    get_log().debug("load_bucket_config", ssm=linz_ssm_bucket_config_name)
```
paulfouquet, line 40: Is there a reason for you writing log messages with underscores `_`?

paulfouquet (approved):
:+1:

topo_processor/util/aws_credentials.py
paulfouquet: `s3_download` is a reference to the function `s3_download()`. Not a big deal, I was just curious :)

blacha: Most of my other libs don't use spaces primarily so I can grep for them without quoting it, I would often use `TileCase`

PR #583, opened 2022-02-15
feat: config for aws stacks TDE-173

amfage:
Like a phoenix from the ashes of my previous two pull requests...
This is to give the stack some more descriptive names, create an S3 bucket, and assign permissions.

### DIFF
--- a/infra/src/batch.ts
+++ b/infra/src/batch.ts
@@ -40,13 +40,13 @@ export class AwsBatchStack extends Stack {
 
     instanceRole.addToPrincipalPolicy(new PolicyStatement({ resources: ['*'], actions: ['sts:AssumeRole'] }));
 
-    const topoProcessorBucket = new Bucket(this, 'TempBucket', {
+    const tempBucket = new Bucket(this, 'TempBucket', {
       removalPolicy: RemovalPolicy.RETAIN,
       blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
       lifecycleRules: [{ expiration: Duration.days(30) }],
     });
 
-    topoProcessorBucket.grantReadWrite(instanceRole);
+    tempBucket.grantReadWrite(instanceRole);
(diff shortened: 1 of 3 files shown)


infra/src/batch.ts
```diff
@@ -39,13 +40,14 @@ export class AwsBatchStack extends Stack {
 
     instanceRole.addToPrincipalPolicy(new PolicyStatement({ resources: ['*'], actions: ['sts:AssumeRole'] }));
 
-    const bucket = new Bucket(this, 'Bucket', {
+    const topoProcessorBucket = new Bucket(this, 'TempBucket', {
```
paulfouquet, line 43: Why not giving the same name for the variable (`const tempBucket = ...`)?

infra/src/submit.ts
```diff
@@ -13,20 +13,18 @@ async function main(): Promise<void> {
   const environment = [
     { name: 'LINZ_CORRELATION_ID', value: correlationId },
     { name: 'LINZ_SSM_BUCKET_CONFIG_NAME', value: 'BucketConfig' },
-    { name: 'LDS_CACHE_BUCKET', value: 'linz-lds-cache' },
+    { name: 'LINZ_CACHE_BUCKET', value: 'linz-lds-cache' },
```
paulfouquet, line 16: Did I miss something and `LDS_CACHE` is now named `LINZ_CACHE`?

paulfouquet (approved):
Looks good. Just a couple of question for my curiosity :)

infra/src/submit.ts
amfage: Oops yes that is from another PR, I've reverted as that PR is not yet merged.

blacha: Ahh I renamed it so everything started with `LINZ_` no real reason just so it looks nice :D

MDavidson17 (approved):

PR #586, opened 2022-02-16
feat: add correlationId to target location TDE-173

amfage:
Questions:

- Would it be better to do this on the Python side?
- Should we be using the correlationId instead of a new ulid for the temp directory?


blacha:
> Would it be better to do this on the Python side?
I think each invocation on the python side should also add a prefix to the bucket

so you would get

`/:correlationId/:jobId/foo.tiff`

so this to me looks good

blacha (approved):

paulfouquet (approved):

PR #653, opened 2022-03-09
fix: change logs for filter_metadata and avoid looping over all metadata records while upload with survey ID (TDE-326)

paulfouquet:
- A logging has been removed as it is not really useful to know if a key exists or not.
- The metadata filtered by survey id is now saved (when `upload` is called with a survey id - not a path) which improve performances as the next search for items in the metadata is done on the filtered records and not the full list of records.


amfage (approved):
LGTM +1:

PR #658, opened 2022-03-10
feat(batch): add geostore roles to temp buckets

MDavidson17:
To deploy locally you will need to add:
`AWS_ORG_ID=****` (where the id is in slack) to your .bashrc file and reload your terminal.


.github/workflows/build.ts.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Build and test
         uses: linz/action-typescript@v1
+        env:
+          AWS_ORG_ID: ${{secrets.AWS_ORG_ID}}
```
blacha, line 18: how come you need this for the build and test phase? does something hit the CDK logic?

blacha (approved):

.github/workflows/build.ts.yml
MDavidson17: Not sure. I was being cautious. I'll test without.

paulfouquet (approved):

PR #679, opened 2022-03-15
feat: use survey name for collection title (TDE-217)

paulfouquet:
For **Historical Imagery**, the `collection title` value is now the survey `name` from **LDS cache layer 51000**.
_Part of this change has been done a refactoring of the `csv_loader` to make it more generic._


topo_processor/util/tests/transfer_collection_test.py
```diff
@@ -76,7 +79,7 @@ def test_generate_summaries(setup: str) -> None:
     metadata_loader_imagery_historic.add_spatial_extent(item_1, asset_metadata=test_geom)
     item_1.datetime = test_datetime
     item_1.properties = {
-        "mission": "SURVEY_1",
+        "mission": "AUCKLAND 1",
```
amfage, line 82: I just clarified with @dwsilk that `mission` should remain as `survey` as it was previously, not the new `title`.

amfage (approved):
:+1:

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -122,14 +122,15 @@ def populate_item(self, metadata_row: Dict[str, str], asset: Optional[Asset] = N
...
-            else:
-                return None
-        else:
-            return survey
+    def get_title(self, survey: str) -> str:
+        survey_names = get_metadata(DataType.SURVEY_FOOTPRINT_HISTORIC)
+        title: str = ""
+
+        if len(survey_names) == 0:
+            raise Exception(f"No name found for survey {survey}")
```
MDavidson17, line 130: Because we're using csvs, is there the possibility that there is no survey name but this is represented with an empty str?
I think the empty str would still pass through here and we'd get a collection with no title.

paulfouquet: Good point @MDavidson17 - I've fixed that now.

MDavidson17 (approved):

PR #713, opened 2022-03-24
feat: sort json keys for historical imagery TDE-338

amfage:
A little bit of re-ordering to tidy things up.


topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -260,5 +250,17 @@ def is_valid_centroid(self, item: Item, centroid: Dict[str, Any]) -> bool:
         return True
 
     def add_linz_geospatial_type(self, item: Item, photo_type: str) -> None:
-
         item.linz_geospatial_type = historical_imagery_photo_type_to_linz_geospatial_type(photo_type)
+
+    def add_stac_extensions(self, item: Item) -> None:
+        item.add_extension(StacExtensions.eo.value)
```
blacha, line 256: Im not sure im sold on these extensions all being grouped down here, to me it makes much more sense to add the extension when their corresponding keys are added

Are any of these extensions only added when the metadata for them is present? 

for instance `eo:bands` should only be added if there is a asset and it has bands? granted all (?) of the historical imagery assets should have a band count?

amfage: I believe extensions should be added anyway. I can move them back into the separate functions.

paulfouquet (changes requested):
As this code will be difficult to maintain (keep the order), it would be good to implement some tests to verify the correct order.

EDIT: @amfage and @dwsilk had a chat about that and it is a temporary order before having tools to manipulate the data in these `json` documents.

paulfouquet (approved):

MDavidson17 (approved):

PR #714, opened 2022-03-24
feat: invoke geostore lambda functions (TDE-331)

paulfouquet:
## Description
This is the implementation of some wrappers to call the **Geostore** _lambda functions_ to be able to upload the processed imagery data (_COGs_ and _STAC_ metadata files) to the **Geostore**.
> **_NOTE:_**  The data to upload to the **Geostore** has to be on a _s3_ bucket.

## Changes
Some CLI commands have been created:
### `add`
The aim of this command is to upload a _survey_ as a _dataset version_ in the **Geostore**.
#### parameters
- `source`: the  _s3_ path to the _survey_ to upload to the **Geostore**. Usually the path to the _survey_ on the _s3_ temporary bucket which has been processed by the **Topo-Processor** `upload` command
- `role`: the _AWS ARN role_ which gives permission to read the `source`. This role will be pass as an argument to the **Geostore** _lambda function_ **datasets** so the **Geostore** can copy the files but also used by the system to get some information related to the survey.
- `commit`: this flag allow the user to actually invoke the data uploading to the **Geostore**. Without this flag, the `add` command only list the files in the `source` path.
#### workflow
1. Download the _collection.json_ **STAC** file in the _survey_ path
2. Get the _ID_ and the _title_ from the _collection.json_
3. If `commit` flag is not specified, log to the user the list of the files which would be upload to the **Geostore** by listing the _path_
4. Check if there is an existing _dataset_ for this _survey_ in the **Geostore** by calling the _datasets lambda function_ with the `GET` method
5. If there is an existing _dataset_ for this _survey_, get the _dataset ID_ and ask the user if she wants to create a new version for it. If user says no (`n`) terminate the process
6. If there is no existing _dataset_ for this _survey_, create one by calling the _datasets lambda function_ with the `POST` method passing the _survey ID_ and `survey title` and get the _dataset ID_ in the _response_
7. Create a new version for the survey by calling the _dataset-version lambda function_ passing the _dataset id_, the `collection.json` in the `source` path and the `role` for the **Geostore** to be able to read the _s3_ bucket. Get the _execution_arn_ from the _response_
8. Call the _import-status lambda function_ with the _execution_arn_ to get a live information of the import status. Give the command to run to get update from the _import-status_ to the user so she can run it later.

### `status`
The aim of this command is to retrieve the upload status of the _survey_ to the **Geostore**.
#### parameter
- `execution-arn`: this information is sent in the _response_ of the invokation of the _dataset-version lambda function_ when creating a new version for a _dataset_. It identify an upload initiated by the **Geostore**. 
#### workflow
The system calls the _import-status lambda function_ with the _execution_arn_ to get a live information of the import status. The system logs the import status received.

### `list`
The aim of this command is to list the versions of a specific or all the _dataset(s)_ in the **Geostore**.
#### parameter
- `survey`: this optional parameter allows to filter the list on the _survey id_
#### workflow
The system calls the _datasets lambda function_ with the `GET` method and specify the `survey` if it was given by the user. Giving the `survey` will get the _versions_ of the `survey` whereas not specifying any `survey` will list all the _versions_ for all the _surveys_. The system returns the output of the _lambda function_ return.

### `delete`
This command had been implemented during development to do some tests. It is not really useful as once a version exists on a _dataset_, you can't delete it by invoking the **Geostore** _lambda function_. You will have to contact the **Geostore** support to do that.

## Usage
`poetry run [COMMAND]`

## Tests
You may want to do some tests on the **Geostore** _nonprod_ environment. To do that you have to modify `topo_processor/geostore/invoke.py` to point to the _nonprod_ environment. The `ROLE_ARN` to use is `arn:aws:iam::632223577832:role/nonprod-api-users` and the function names take the prefix _nonprod-_: `name = "nonprod-" + name`

## Sources
https://github.com/linz/geostore/blob/master/USAGE.md
https://confluence.linz.govt.nz/display/TOP/Uploading+to+The+Geostore


topo_processor/cli/geostore/delete.py
```diff
@@ -0,0 +1,54 @@
...
+def main(datasetid: str, prod: str, verbose: str) -> None:
+    start_time = time_in_ms()
+    get_log().info("delete_datasets_start", dataset_id=id)
+
+    if verbose:
+        set_level(LogLevel.trace)
+
+    client = boto3.client("lambda")
+
+    if prod:
```
paulfouquet, line 37: Probably a better way to do that?

topo_processor/cli/geostore/delete.py
```diff
@@ -0,0 +1,54 @@
...
+    is_flag=True,
+    help="Use this flag to export into production environment.",
+)
+@click.option(
+    "-v",
+    "--verbose",
+    is_flag=True,
+    help="Use verbose to display trace logs",
+)
+def main(datasetid: str, prod: str, verbose: str) -> None:
```
paulfouquet, line 28: The different `geostore` wrapped function (`delete`, `export`, `list`, `status`) have some duplicate code. Any thoughts?

topo_processor/geostore/invoke.py
```diff
@@ -0,0 +1,41 @@
...
+from linz_logger import get_log
+
+if TYPE_CHECKING:
+    from mypy_boto3_lambda import Client
+else:
+    Client = object
+
+
+def invoke_lambda(client: Client, name: str, http_method: str, parameters: Dict[str, str]) -> Dict[str, Any]:
+    payload = b'{"http_method": "' + http_method.encode() + b'", "body": {'
```
blacha, line 13: json.dumps

topo_processor/cli/geostore/export.py
```diff
@@ -0,0 +1,96 @@
...
+    if prod:
+        environment = "prod"
+    else:
+        environment = "nonprod"
+
+    try:
+
+        # create a dataset
+        create_dataset_parameters = {"title": surveyid, "description": surveytitle}
+        dataset_response_payload = invoke_lambda(client, f"{environment}-datasets", "POST", create_dataset_parameters)
```
blacha, line 69: in prod these will not have `${env}-datasets` eg `datasets` not `prod-datasets`

pyproject.toml
```diff
@@ -32,6 +32,13 @@ pytest-cov = "^3.0.0"
 pytest-mock = "^3.6.1"
 mypy = "*"
 
+[tool.poetry.scripts]
+export = 'topo_processor.cli.geostore.export:main'
+status = 'topo_processor.cli.geostore.status:main'
+list = 'topo_processor.cli.geostore.list:main'
+delete = 'topo_processor.cli.geostore.delete:main'
+validate = 'topo_processor.cli.validate:main'
+
```
amfage, line 41: I'm not generally a fan of typing more than necessary but I wonder if it would be clearer if the commands were prefixed by `geostore-` (or something else more identifiable/less generic)?

topo_processor/cli/geostore/delete.py
amfage: Could it be one command e.g. `geostore-upload` and then a mode option i.e. `add`, `delete`, `list`, `status` followed by the specific flags for each mode e.g.  `--datasetid`.
Then some of the code could be reused as well.

paulfouquet: I had thought about that before but I did not want to have more arguments that there are already

MDavidson17 (approved):

topo_processor/cli/geostore/delete.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if verbose:
+        set_level(LogLevel.trace)
+
+    client = boto3.client("lambda")
+
+    try:
+        delete_parameters = {"id": dataset_id}
+        response = invoke_lambda(client, "datasets", "DELETE", delete_parameters, prod)
+        if response["status_code"] != 204:
```
blacha, line 40: generally checking for a specific 200 status code is not a good idea, is there some sort of `response.ok`? which would be `status_code >=200 && <= 299`

topo_processor/cli/geostore/delete.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@click.command()
+@click.option(
+    "-id",
+    "--dataset-id",
+    required=True,
+    help="The dataset id to delete",
+)
+@click.option(
```
blacha, line 16: its a good idea to have either a `--dry-run` or default to dry run and have a `--commit` before deleting things,

`delete.py  --dataset-id 12345 --commit` to me makes more sense than a  `delete.py --dataset-id 12345 -p`

topo_processor/cli/geostore/export.py
```diff
@@ -0,0 +1,88 @@
...
+
+@click.command()
+@click.option(
+    "-s",
+    "--source",
+    required=True,
+    help="The s3 path to the collection.json of the survey to export",
+)
+@click.option(
+    "-sid",
```
blacha, line 20: how does this work? How does this handle `-s -i -d` ? 

generally these should be a single letter?

topo_processor/geostore/invoke.py
```diff
@@ -0,0 +1,50 @@
...
+if TYPE_CHECKING:
+    from mypy_boto3_lambda import Client
+else:
+    Client = object
+
+
+def invoke_lambda(
+    client: Client, name: str, http_method: str, parameters: Dict[str, str], is_prod: bool = False
+) -> Dict[str, Any]:
+    if not is_prod:
```
blacha, line 15: I dont think we should be hitting any "nonprod" environment that is not ours

paulfouquet: That's a good point. I've been using their "nonprod" environment to do my tests during development. Now, there is no reason to use this "nonprod" environment. Let's include your option of a `--commit` flag to have a sort of confirmation for any `add`/`delete`

dwsilk: It's not their dev environment, it's a clone of prod for us to use for test uploads.

topo_processor/cli/geostore/delete.py
paulfouquet: The response when a delete is successful is `{"status_code": 204, "body": {}}` so I don't know what to do with that?

blacha: I would make a is_ok function that checks that status is in the 200 range

topo_processor/geostore/invoke.py
paulfouquet: Another thing to note is that `-p` or `--prod` works only if you're connected to the AWS Geostore prod account. And non-prod (default mode) works only if you're connected to non-prod. The issue is that the lambda function name are different if you use prod or non-prod. Maybe we could check in the system if the user is authenticated to prod or non prod to identify which function name to use? That way, we get rid of the `-p` or `--prod` flag, and maybe use the `--commit` mode for `add` too?

topo_processor/cli/geostore/list.py
```diff
@@ -0,0 +1,36 @@
...
+    required=False,
+    help="The survey to filter",
+)
+@click.option(
+    "-v",
+    "--verbose",
+    is_flag=True,
+    help="Use verbose to display debug logs",
+)
+def main(dataset_id: str, verbose: bool) -> None:
```
l0b0, line 21: Is a mapping from `--survey` to `dataset_id` intentional, or should one of them be renamed?

topo_processor/cli/geostore/list.py
```diff
@@ -0,0 +1,36 @@
...
+    start_time = time_in_ms()
+    get_log().info("list_datasets_start", dataset_id=dataset_id)
```
l0b0, line 23: Is running `get_log()` for every log occurrence intentional? Would it be possible to pull out a constant/variable instead? Ditto everywhere there's more than one `get_log()` call in a single context.

topo_processor/geostore/tests/invoke_test.py
```diff
@@ -0,0 +1,34 @@
...
+        "status_code": 200,
+        "body": {
+            "created_at": "2022-03-23T02:41:53.940795+0000",
+            "pk": "DATASET#01FYTADW8MSCNR8D68EX7APMD3",
+            "title": "test_title",
+            "updated_at": "2022-03-23T02:41:53.940911+0000",
+            "id": "01FYTADW8MSCNR8D68EX7APMD3",
+        },
+    }
+    assert is_response_ok(response_ok) is True
```
l0b0, line 34: `is True` is redundant.

README.md
```diff
@@ -129,7 +129,68 @@ The following source and target combinations can be used:
 | local  | local  |
 | local  |   s3   |
 
-### Validate
+**For the following Geostore commands (described with `(Geostore)`), you need to be authenticated into a Geostore AWS account, assuming the `nonprod-api-users` role. These commands have a `-p` or `-prod` flag. You won't be able to use the `-prod` flag if you authenticated in a `Geostore` `non-prod` account and vice versa. **
```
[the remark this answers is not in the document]
paulfouquet: Good spotting, I forgot to fix that in the README

pyproject.toml
```diff
@@ -23,14 +23,23 @@ python-dotenv = "^0.20.0"
...
 mypy = "*"
+moto = "^3.1.4"
+mypy-boto3-lambda = "^1.24.0"
+
+[tool.poetry.scripts]
+add = 'topo_processor.cli.geostore.add:main'
+status = 'topo_processor.cli.geostore.status:main'
+list = 'topo_processor.cli.geostore.list:main'
+delete = 'topo_processor.cli.geostore.delete:main'
+validate = 'topo_processor.cli.validate:main'
```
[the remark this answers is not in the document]
l0b0: If the PR is going to be squashed I would definitely do that. Otherwise, meh. It's already been reviewed.

README.md
```diff
@@ -129,7 +129,68 @@ The following source and target combinations can be used:
...
+This is to follow the current upload status to the `Geostore` for a particular `dataset` version. You may have to run it several times as the status gets updated.
+
+| Argument                   |                              Description                              |
+| -------------------------- | :-------------------------------------------------------------------: |
+| `-a` or `--execution-arn`  | The execution arn received from the Geostore after invoking an upload |
+| `-v` or `--verbose` (FLAG) |                   Use this flag display debug logs                    |
+
+**_NOTE:_** The command to run is given in the logs after calling sucessfully the `add` command:
+
+```json
```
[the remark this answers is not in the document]
paulfouquet: I might remove this

topo_processor/cli/geostore/add.py
```diff
@@ -0,0 +1,155 @@
...
+                    assumed_role["Credentials"]["AccessKeyId"],
+                    assumed_role["Credentials"]["SecretAccessKey"],
+                    assumed_role["Credentials"]["SessionToken"],
+                )
+                s3_download(os.path.join(source, "collection.json"), collection_local_path, credentials)
+            except Exception as e:
+                get_log().error("geostore_export_failed", source=source, error=e)
+                return
+        else:
+            raise Exception("The source has to be a survey in a S3 bucket.")
```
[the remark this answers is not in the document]
paulfouquet: I understand and it's something I would have done if we wanted to handle those exceptions in a specific way. Currently we don't so we haven't create any `Exception` subclasses. If we do have a specific need in the future, then we'll refactor that. Is that a good approach? Or you'll suggest to separate things now even if there is no specific need at the moment (and probably not in the future?)

topo_processor/cli/geostore/add.py
```diff
@@ -0,0 +1,155 @@
...
+        set_level(LogLevel.info)
+
+    try:
+        source_role_arn = role
+        client_sts = boto3.client("sts")
+        assumed_role = client_sts.assume_role(RoleArn=source_role_arn, RoleSessionName="read-session")
+        # Get Collection information
+        collection_local_path = os.path.join(temp_folder, "collection.json")
+
+        if is_s3_path(source):
```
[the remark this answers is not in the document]
paulfouquet: Good point

topo_processor/cli/geostore/list.py
paulfouquet: Not intentional, I forgot to change `dataset_id` to `survey` after making the change to call the lambda with the survey `title`

topo_processor/cli/geostore/list.py
paulfouquet: No, good point. As we do that everywhere in the code, I'll rather create another Pull Request to fix that everywhere

README.md
```diff
@@ -129,7 +129,68 @@ The following source and target combinations can be used:
...
-### Validate
+**For the following Geostore commands (described with `(Geostore)`), you need to be authenticated into a Geostore AWS account, assuming the `nonprod-api-users` role. These commands have a `-p` or `-prod` flag. You won't be able to use the `-prod` flag if you authenticated in a `Geostore` `non-prod` account and vice versa. **
+
+### `add` (Geostore)
+
+This command allows to add a survey to the [Geostore](https://github.com/linz/geostore) by using the [Geostore API](https://github.com/linz/geostore/blob/master/USAGE.md).
+
+**_Prerequisites:_** The survey has to be processed by the `upload` command first. The output files of the `upload` is what will be exported to the `Geostore`.
+
+| Argument                   |                     Description                     |
```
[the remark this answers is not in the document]
paulfouquet: Thanks for the tips using `--help`

README.md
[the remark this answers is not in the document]
paulfouquet: It is actually properly formatted.

topo_processor/cli/geostore/add.py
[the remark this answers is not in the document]
l0b0: OK, fair enough. It's not a lot of tech debt.

topo_processor/cli/validate.py
```diff
@@ -24,6 +24,7 @@
     help="Use collection to validate collections only.",
 )
 @click.option(
+    "--m",
```
amfage, line 27: `-m`

README.md
```diff
@@ -129,7 +129,66 @@ The following source and target combinations can be used:
...
+| `-s`, `--survey` TEXT |       The survey to filter        |
+| `-v`, `--verbose`     | Use verbose to display debug logs |
+
+```bash
+poetry run list [-s ID123ABC]
+```
+
+### `delete` (Geostore)
+
+Delete a dataset into the `Geostore`. Only if the dataset does not contain any version. To delete a dataset which contains a version, contact the **Geostore** support.
```
amfage, line 179: Minor point, it should be "from" instead of "into".

amfage (approved):

PR #728, opened 2022-03-28
feat(cdk): lambda function to monitor batch job failures

MDavidson17:
Creates a lambda function to log to elastic search events in `li-topo-prod` and `li-topo-nonprod` that follow the pattern:
```
source: ['aws.batch'],
detailType: ['Batch Job State Change'],
detail: { status: ['FAILED'] },
```

**nb: I added `--all` to the nonprod and prod deployment workflows in `.github/workflows/build.ts.yml` this means it should deploy both `TopoProcessorBatch` and `TopoProcessorMon` on merge to master for nonprod and on publish for prod.** - Is this ok?

### DIFF
--- a/infra/src/batch-monitor.ts
+++ b/infra/src/batch-monitor.ts
@@ -1,0 +1,29 @@
+import { Stack, StackProps } from 'aws-cdk-lib';
+import * as events from 'aws-cdk-lib/aws-events';
+import * as evtTargets from 'aws-cdk-lib/aws-events-targets';
+import * as lf from 'aws-cdk-lib/aws-lambda';
+import { Code } from 'aws-cdk-lib/aws-lambda';
+import { Construct } from 'constructs';
+import * as path from 'path';
+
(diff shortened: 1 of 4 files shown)


infra/src/batch-monitor.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class AwsBatchMonitor extends Stack {
+  public constructor(scope: Construct, id: string, props: StackProps) {
+    super(scope, id, props);
+
+    const rule = new events.Rule(this, 'BatchEventRule', {
+      eventPattern: {
+        source: ['aws.batch'],
+        detailType: ['Batch Job State Change'],
+        detail: { status: ['FAILED'] },
```
blacha, line 17: It might be worth it to just track all status changes? its super cheap to trace them all ? and might give us interesting information sometime in the future?

paulfouquet (approved):

amfage (approved):

PR #750, opened 2022-03-31
feat: log unique values

MDavidson17:
- Checksum on file transferred
- collection.id on stac collection object created
- item.id on stac item object created


amfage (approved):

paulfouquet (approved):
:+1:

PR #764, opened 2022-04-04
feat: iterate all the extensions (warnings instead of raising Exception)

paulfouquet:


amfage (dismissed):
:+1:

topo_processor/metadata/metadata_validators/metadata_validator_stac.py
```diff
@@ -31,11 +32,15 @@ def is_applicable(self, stac_object: Union[Item, Collection]) -> bool:
         return True
 
     def validate_metadata(self, item: Item) -> None:
-        try:
+
+        with warnings.catch_warnings(record=True) as w:
```
paulfouquet, line 36: Should I check here if the `IterErrorsValidator` is the one used? Otherwise that won't match with the default validator (which does not raise `warnings` but `exception`). What the best way to do that? a global variable set to `True` when the custom validator is enabled? (`pystac.validation.set_validator(IterErrorsValidator())`

amfage: Good point. Are we ever going to use the default validator? I guess it would be safest to check. How would this behave with jsonschema-rs (will we need to use that in future?).

paulfouquet: If we want to use the custom validator by default, something needs to be done to force it all the time (instead of setting it for `upload` as it is currently done. 

> How would this behave with jsonschema-rs (will we need to use that in future?)

Only the future will tell us :) I guess we will make a change when/if this will be needed.

conftest.py
```diff
@@ -7,7 +7,7 @@
 
 from topo_processor.stac.iter_errors_validator import IterErrorsValidator
 
-pystac.validation.set_validator(IterErrorsValidator())
+# pystac.validation.set_validator(IterErrorsValidator())
```
paulfouquet, line 10: Why is this commented?

amfage: It was for testing, I'm just updating it.

conftest.py
```diff
@@ -7,7 +7,7 @@
 
 from topo_processor.stac.iter_errors_validator import IterErrorsValidator
 
-# pystac.validation.set_validator(IterErrorsValidator())
+pystac.validation.set_validator(IterErrorsValidator())
```
paulfouquet, line 10: Using a `fixture` with the `autouse=True` (`@pytest.fixture(autouse=True)`) will set this validator before each test.
```suggestion
@pytest.fixture(autouse=True)
def set_iter_errors_validator() -> None:
 pystac.validation.set_validator(IterErrorsValidator())
```

Check out [Autouse fixture](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request)

When you need to use another validator you could hardcode it at the beginning of the test.
That would avoid you to set back the `IterErrorsValidator` at the end of the `test_check_multiple_stac_extensions_default_pystac_validator()` test too.

topo_processor/metadata/metadata_validators/tests/metadata_validator_stac_test.py
```diff
@@ -141,6 +143,61 @@ def test_check_validity_version_extension() -> None:
...
+    item.add_extension(StacExtensions.camera.value, add_to_collection=False)
+    validator = MetadataValidatorStac()
+    assert isinstance(
+        pystac.validation.RegisteredValidator.get_validator(), pystac.validation.stac_validator.JsonSchemaSTACValidator
+    )
+    assert validator.is_applicable(item)
+    with pytest.raises(STACValidationError) as e:
+        validator.validate_metadata(item)
+    assert "aerial-photo" in str(e.value)
+    assert not "camera" in str(e.value)
```
paulfouquet, line 171: I like that we can easily understand the difference

amfage (approved):

MDavidson17 (approved):
:+1:

PR #773, opened 2022-04-06
fix: update quality-description and collection-description

MDavidson17:


topo_processor/stac/collection.py
```diff
@@ -49,6 +50,7 @@ def __init__(self, title: str):
         # FIXME: Do we want to generate this id like this?
         self.id = str(ulid.ULID())
         self.title = title
+        self.description = "default description"
```
paulfouquet, line 53: Why is this needed? (instead of initializing with an empty string)

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -121,6 +107,22 @@ def populate_item(self, metadata_row: Dict[str, str], asset: Optional[Asset] = N
         item.add_extension(StacExtensions.linz.value)
         item.add_extension(StacExtensions.version.value)
 
+    def population_collection(self, collection: Collection) -> None:
+        collection.description = "This aerial photographic survey was digitised from {} {} negatives in the Crown collection of the Crown Aerial Film Archive."
```
paulfouquet, line 111: Could this be all done in `collection.update_description()`? I get that this will be updated with the `colour` and `size` later but for understanding I would suggest to do everything at the same place.

paulfouquet (commented):
That looks good. Just have a couple of questions

amfage (approved):

topo_processor/metadata/metadata_loaders/metadata_loader_imagery_historic.py
```diff
@@ -121,6 +107,21 @@ def populate_item(self, metadata_row: Dict[str, str], asset: Optional[Asset] = N
         item.add_extension(StacExtensions.linz.value)
         item.add_extension(StacExtensions.version.value)
 
+    def population_collection(self, collection: Collection) -> None:
```
paulfouquet, line 110: As your function change the state of the `collection`, it's more clear to use a verb (i.e. `populate` instead of `population`) for the function name.

paulfouquet (changes requested):
As @Megan is away I'll commit this quick change on her behalf.

paulfouquet (approved):

PR #935, opened 2022-05-23
fix: process should stop and files not transfered if any validation fails (TDE-354)

paulfouquet:
### Fixes: 
_The upload process is now aborted if at least one `Item` or the `Collection` validation fails._
_The files are transfered at the end of the process (after the validations) so if something fails during validation, the files are not transfered._

### New:
_Force mode can be enabled with flag `--force`. It allows the upload even if some data are invalid. For example, if the `Collection` validation fails or if some `Items` are excluded because they are invalid (the invalid ones won't be uploaded)._


amfage (approved):
:+1:

blacha (approved):

PR #1013, opened 2022-06-13
fix: TDE-392 stop processing when an image transform fails

amfage:
This also removes a noisy logging line in the LDS cache code that was polluting our ElasticSearch logging.

### DIFF
--- a/topo_processor/data/data_transformers/data_transformer_imagery_historic.py
+++ b/topo_processor/data/data_transformers/data_transformer_imagery_historic.py
@@ -38,7 +38,12 @@ class DataTransformerImageryHistoric(DataTransformer):
                 return
             output_path = os.path.join(item.collection.get_temp_dir(), f"{ulid.ULID()}.tiff")
 
-            create_cog(asset.source_path, output_path).run()
+            try:
+                test = create_cog(asset.source_path, output_path).run()
+            except Exception as e:
+                raise Exception(
+                    f"COG creation failed for item {item.id} with source path {asset.source_path} and output path {output_path}. Process is stopped."
+                ) from e
 
             get_log().debug("Created COG", output_path=output_path, duration=time_in_ms() - start_time)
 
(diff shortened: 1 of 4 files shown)


MDavidson17 (approved):
:+1:

topo_processor/data/data_transformers/data_transformer_imagery_historic.py
```diff
@@ -38,7 +38,12 @@ def transform_data(self, item: Item) -> None:
                 return
             output_path = os.path.join(item.collection.get_temp_dir(), f"{ulid.ULID()}.tiff")
 
-            create_cog(asset.source_path, output_path).run()
+            try:
+                test = create_cog(asset.source_path, output_path).run()
```
paulfouquet, line 42: If there is no use for the variable `test` I would suggest to not declare it

topo_processor/data/data_transformers/data_transformer_repo.py
```diff
@@ -26,8 +26,8 @@ def transform_data(self, item: Item) -> None:
                     transformer.transform_data(item)
                 except Exception as e:
                     item.add_error(str(e), transformer.name, e)
-                    get_log().warning(f"Data Transform Failed: {e}", transformers=transformer.name)
-                    return
+                    get_log().error(f"Data Transform Failed: {e}", transformers=transformer.name)
```
paulfouquet, line 29: I wonder if it is the error is not display in a better way (specially in Elastic Search) if you declare it as an object like:
```
get_log().error(f"Data Transform Failed", transformers=transformer.name, error[or err]=e)
```

paulfouquet (changes requested):
The logic looks good. Just a couple of suggestions.

topo_processor/data/data_transformers/data_transformer_imagery_historic.py
amfage: Oops left over from testing, thank you.

paulfouquet (approved):

PR #1039, opened 2022-06-16
fix: TDE-402 GitHub actions git tag order

amfage:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM osgeo/gdal:ubuntu-small-latest
+FROM osgeo/gdal:ubuntu-small-3.5.0
```
paulfouquet, line 1: I think this is not part of this PR?

paulfouquet (approved):

Dockerfile
amfage: Noo it is not.

paulfouquet: It will get merged with the other one anyway :)

PR #1054, opened 2022-06-22
fix: TDE-405 geostore title change

amfage:


topo_processor/cli/geostore/add.py
```diff
@@ -42,15 +50,21 @@
...
     """Create or add a new version of an existing dataset to the Geostore for the source (survey) passed as argument."""
     start_time = time_in_ms()
+    data_type = DataType(datatype)
     logger = get_log()
     logger.info("geostore_add_started", source=source)
 
     if not verbose:
         set_level(LogLevel.info)
 
+    if data_type == DataType.IMAGERY_HISTORIC:
```
paulfouquet, line 63: That's a shame that there is no information about the data type in the STAC metadata. Or I missed it?
It would be more convenient to get this information from the `collection.json` so the user does not have to provide it in the command.
Without having this possibility, the approach you took would be the on I'd take too...

README.md
```diff
@@ -165,10 +166,11 @@ This is to follow the current upload status to the `Geostore` for a particular `
...
 It gives you the information for one or all the datasets created on the `Geostore`.
 
-| Argument              |            Description            |
-| --------------------- | :-------------------------------: |
-| `-s`, `--survey` TEXT |       The survey to filter        |
-| `-v`, `--verbose`     | Use verbose to display debug logs |
+| Argument                |                                Description                                |
+| ----------------------- | :-----------------------------------------------------------------------: |
+| `-s`, `--survey` TEXT   |                           The survey to filter                            |
+| `-d`, `--datatype` TEXT | Datatype _Only `imagery.historic` is available at the moment._ [required] |
```
paulfouquet, line 172: To keep the documentation aligned with the CLI help, you can use `poetry run [command] --help` and copy paste the return in the documentation.

topo_processor/cli/geostore/add.py
amfage: We could infer it from the collection.json e.g. the `stac_extensions` :thinking:

paulfouquet: Yes I thought about that too but that looks too "guessing"

paulfouquet (approved):

dwsilk (approved):

PR #1153, opened 2022-07-19
fix: TDE-410 LDS cache file convert geopackage to CSV

amfage:
This is a fix to convert gpkg files from the LDS to csv format pending a longer term plan for properly handling geopackages.


topo_processor/metadata/lds_cache/lds_cache.py
```diff
@@ -51,11 +53,31 @@ def get_metadata(
...
             else:
                 raise Exception(f"{metadata_path} not found")
 
     if os.path.isfile(metadata_path):
+        # This doesn't handle gzipped files as input on the command line
+        if is_geopackage(metadata_path):
+            new_metadata_path = os.path.splitext(metadata_path)[0] + ".csv"
+            if os.path.isfile(new_metadata_path):
+                raise Exception(
+                    f"Conversion of gpkg file {metadata_path} will overwrite existing csv file {new_metadata_path}"
```
paulfouquet, line 74: Did you want to stop the process at this point? I have the feeling that the message is a warning message rather than an error. I also don't see it as a problem to overwrite it as it is generated from the LDS cache.

topo_processor/metadata/lds_cache/lds_cache.py
```diff
@@ -51,11 +53,31 @@ def get_metadata(
...
+                        decompress_file(metadata_path, True)
+                        new_metadata_path = os.path.splitext(metadata_path)[0] + ".csv"
+                        geopackage_to_csv(metadata_path, new_metadata_path).run()
+                        metadata_path = new_metadata_path
+                    else:
+                        raise Exception(f"Unsupported file format. {metadata_path} must be .csv or .gpkg")
             else:
                 raise Exception(f"{metadata_path} not found")
 
     if os.path.isfile(metadata_path):
```
paulfouquet, line 68: Do you think that we can avoid some code duplication here?

paulfouquet (commented):
A couple of questions

topo_processor/metadata/lds_cache/lds_cache.py
```diff
@@ -52,10 +55,24 @@ def get_metadata(
...
                 if exported_asset.extra_fields.get("encoding", None) == "gzip":
                     decompress_file(metadata_path)
+                    new_metadata_path = os.path.splitext(metadata_path)[0] + "_" + time.strftime("%s") + ".csv"
+                    geopackage_to_csv(metadata_path, new_metadata_path).run()
+                    metadata_path = new_metadata_path
             else:
                 raise Exception(f"{metadata_path} not found")
 
     if os.path.isfile(metadata_path):
+        if is_geopackage(metadata_path):
```
paulfouquet, line 65: There is a code duplication you could get rid of by having
```
if os.path.isfile(metadata_path):
 if exported_asset.extra_fields.get("encoding", None) == "gzip":
   decompress_file(metadata_path)
 new_metadata_path = os.path.splitext(metadata_path)[0] + "_" + time.strftime("%s") + ".csv"
 geopackage_to_csv(metadata_path, new_metadata_path).run()
 metadata_path = new_metadata_path
```
only once

topo_processor/metadata/lds_cache/lds_cache.py
```diff
@@ -52,10 +55,24 @@ def get_metadata(
             if os.path.isfile(metadata_path):
                 if exported_asset.extra_fields.get("encoding", None) == "gzip":
                     decompress_file(metadata_path)
+                    new_metadata_path = os.path.splitext(metadata_path)[0] + "_" + time.strftime("%s") + ".csv"
```
paulfouquet, line 58: That will be executed only if the file is compressed. I understand that it will be executed later, from line 65 because the code is duplicated. You can simplify that by either only `decompress_file` here. In my opinion it's better to do everything at the same place and avoid this code duplication.

topo_processor/util/file_converter.py
```diff
@@ -0,0 +1,18 @@
+import os
+
+from topo_processor.cog.command import Command
```
paulfouquet, line 3: The `command` module is not `COG` specific anymore. I would move it to somewhere more generic (a `system` utils for example).

topo_processor/metadata/lds_cache/lds_cache.py
amfage: Oops my bad that was a mistake.

paulfouquet (approved):

